PR #14, opened 2016-04-18
Kernel nanny proposal

takluyver:
As discussed at the dev meeting. There are a few TODOs which have not yet been decided. We can bikeshed about them, or whoever gets to implementing the relevant bits first can try their preferred option ;-).

Pinging @JanSchulz, who was interested in this for IRkernel logging.

### DIFF
--- a/kernel-nanny.md
+++ b/kernel-nanny.md
@@ -1,0 +1,117 @@
+# Kernel 'nanny' processes
+
+## Summary
+
+We propose to start Jupyter kernels through a 'nanny' process, which will always
+be running on the same machine as its associated kernel. This offers various
+advantages over the current situation, including:
+
+- Kernels will no longer need to implement the 'heartbeat' for frontends to
+  check that they are still alive.
+- We will be able to interrupt remote kernels (SIGINT cannot be sent over the network)
+- There will be a consistent way to start kernels without a frontend
+  (`jupyter kernel --kernel x`).
+- Kernel stdout & stderr can be captured at the OS level, with real-time updates
+  of output.
+
+## The basics
+
+When a frontend wants to start a kernel, it currently instantiates a `KernelManager`
+object which reads the kernelspec to find how to start the kernel, writes a
+connection file, and launches the kernel process. With this process, it will
+instead launch the kernel nanny on the machine where the kernel is to run, and
+the nanny will be responsible for creating the connection file and launching
+the kernel process.
+
+**Rejected alternative:** One kernel nanny process per machine, able to start
+multiple kernels. This would be more complex, but we may come back to it later
+if the overhead of one nanny per kernel is too much.
+
+## Socket connections
+
+Currently, the frontend connects to five sockets to communicate with the kernel:
+
+* Shell
+* Control (priority, used for shutdown)
+* Iopub (kernel to frontend only, for output)
+* Stdin (used to request input from the frontend)
+* Heartbeat
+
+Of these, shell and stdin will remain connected directly between the kernel and
+the frontend. Control and iopub (see output capturing) will be connected through
+the nanny, i.e. each channel will have one socket for communications between
+the frontend and the nanny, and a second socket for communications between the
+nanny and the kernel. (*TODO: What are these called in the connection file? Or
+do we have two connection files?*) The heartbeat will only be between the
+frontend and the nanny, to detect situations such as network failures.
+
+## Messaging changes
+
+* A new message type on the control channel from the frontend to the nanny,
+  instructing the nanny to shut down the kernel.
+* A new message type on the control channel from the frontend to the nanny,
+  instructing the nanny to signal/interrupt the kernel. (*TODO: Expose all Unix
+  signals, or just SIGINT?*)
+* Heartbeat becomes a broadcast signal from the nanny to all connected frontends,
+  rather than a REP/REQ pattern (which was only the case before because pyzmq
+  makes it easy to echo messages without grabbing the GIL).
+* New broadcast message from nanny to frontends when kernel dies unexpectedly,
+  including exit status.
+* New end of output message, from nanny to frontends? (*TODO: yes/no?*)
+
+## Output capturing
+
+In IPython, we capture stdout/stderr at the Python level (sys.std*). Code which
+writes to stdout/stderr at a lower level (e.g. C extensions) will send its output
+to the terminal where the frontend was started, instead of to the frontend.
+Many other kernels suffer from similar issues.
+
+We know of tricks using `dup2` to redirect the low-level file handles within the
+kernel, but we don't want each kernel to reimplement this, and it is not
+possible on Windows.
+
+To this end, when the kernel nanny starts the kernel, it will be able to create
+stdout and stderr as pipes, and turn data read from them into *stream* messages
+to be sent to the frontend via the iopub channel. However, this may make
+debugging difficult or show unwanted output if kernel authors are using the
+terminal to debug the kernel implementation. Therefore, output capturing will
+only be enabled if the kernel opts in via its `kernel.json` specification:
+
+    "capture_stdstreams": true
+
+### Output synchronisation
+
+With this proposal, there are multiple asynchronous channels for output coming
+from the kernel: the `iopub` socket from the kernel to the nanny, and the pipes
+carrying stdout and stderr. At present, the `status` message with
+`execution_state: idle` marks the end of output on the iopub channel.
+
+Kernels that opt in to output capturing should print a delimiter (*TODO: define
+delimiter*) on each of stdout and stderr, before and after running user code.
+The delimiter will include the message ID of the execute_request message.
+The nanny will not forward these to the frontend, but will use the 'before'
+delimiters to indicate which execution output resulted from, and the 'after'
+delimiters to detect when stream output is finished. The nanny will tell the
+frontend when all output from an execute_request, on iopub and the two pipes,
+is complete (*TODO: using status:idle, or a new message?*).
+
+**Rejected alternative:** Frontends monitoring when output is complete on each
+channel. The frontend output handling logic would have to know whether the
+kernel in use used output capturing or not, and the logic would have to be
+written for each frontend. With the scheme we decided upon, the logic must only
+be implemented once in the nanny process, and the frontend can remain ignorant
+of whether the kernel has enabled output capturing.
+
+### Kernel logging
+
+Where kernels have been previously using low-level stdout/stderr to log to the
+terminal, they need a new way to produce diagnostic logs which shouldn't be
+displayed in the frontend. Kernels opting in to output capturing will be
+started with an environment variable `JUPYTER_KERNEL_LOG` set. The kernel
+should treat this as a filesystem path, which it can open and write logs to.
+
+Kernels should make minimal assumptions about the type of file they are opening.
+It may be a regular file, a FIFO (or named pipe), or the slave end of the tty
+where the frontend was started, on systems where that is possible. This will
+likely depend on configuration settings in the frontend, and possibly on how the
+frontend is started. It should never be a directory, however.
--- a/kernel-nanny.md
+++ b/kernel-nanny.md
@@ -18,7 +18,7 @@ advantages over the current situation, including:
 
 When a frontend wants to start a kernel, it currently instantiates a `KernelManager`
 object which reads the kernelspec to find how to start the kernel, writes a
-connection file, and launches the kernel process. With this process, it will
+connection file, and launches the kernel process. With the proposed changes, it will
 instead launch the kernel nanny on the machine where the kernel is to run, and
 the nanny will be responsible for creating the connection file and launching
 the kernel process.


janschulz:
Not sure if I understand the details, but currently the notebook isn't very good at shutting down the R kernel on windows, because the R kernel is not a single process, but more like R.exe -> cmd -> rterm.exe. I'm not sure if the "nanny" can detect such a thing without a heartbeat?

[Such things might happen even for python kernels, if you use a batch file with `activate <env> & python <kernel startup line>`, which is needed to get the correct PATH in a kernel...]

takluyver:
I suspect it won't make much of a difference either way in that situation. Both currently and with the kernel nanny, it will send `shutdown_request` to ask the kernel to shut itself down, and if it doesn't shut down within some time period, it will terminate it more forcefully. I'd guess that second bit is where it goes wrong, since it only knows about the top-level process that it started.

Besides fiddling with the time we wait for the kernel to shut itself down, I'm not sure what we could do to improve that.

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+do we have two connection files?*) The heartbeat will only be between the
+frontend and the nanny, to detect situations such as network failures.
+
+## Messaging changes
+
+* A new message type on the control channel from the frontend to the nanny,
+  instructing the nanny to shut down the kernel.
+* A new message type on the control channel from the frontend to the nanny,
+  instructing the nanny to signal/interrupt the kernel. (*TODO: Expose all Unix
+  signals, or just SIGINT?*)
```
minrk: A `signal_request` message makes the most sense. I don't think there's a reason to limit to interrupt/term/kill, all of which we probably want.

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+possible on Windows.
+
+To this end, when the kernel nanny starts the kernel, it will be able to create
+stdout and stderr as pipes, and turn data read from them into *stream* messages
+to be sent to the frontend via the iopub channel. However, this may make
+debugging difficult or show unwanted output if kernel authors are using the
+terminal to debug the kernel implementation. Therefore, output capturing will
+only be enabled if the kernel opts in via its `kernel.json` specification:
+
+    "capture_stdstreams": true
```
minrk: A problem that this introduces that we should address is that now kernels cannot write to the terminal at all - there is no way for kernels to have logging without going straight to a file.

minrk: (duh, I should finish reading, I see `# Kernel logging` two seconds down)

minrk:
Thanks for writing this up, @takluyver!

kernel-nanny.md
takluyver: For Unix systems that certainly makes sense. For Windows, should we just pick some numbers to refer to the available ways we have of interrupting/stopping the kernel process?

minrk: I think only one or two signals work on Windows reliably, but they are still integers, aren't they?

takluyver: AIUI Windows doesn't really have signals at all, but Python exposes certain similar operations through the same interface it uses for signals on Windows. The description of os.kill has some useful info:

https://docs.python.org/3/library/os.html#os.kill

We could quite reasonably expose the same set of options with the same meanings as Python does, of course.

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+- There will be a consistent way to start kernels without a frontend
+  (`jupyter kernel --kernel x`).
+- Kernel stdout & stderr can be captured at the OS level, with real-time updates
+  of output.
+
+## The basics
+
+When a frontend wants to start a kernel, it currently instantiates a `KernelManager`
+object which reads the kernelspec to find how to start the kernel, writes a
+connection file, and launches the kernel process. With this process, it will
```
willingc: Perhaps for clarity:
"With this proposed process, the frontend will..."

willingc:
@takluyver Nicely designed and clearly written. I made a drawing by hand of the frontends, nanny, kernel, and channels; let me know if you would like a copy. 😄

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+be running on the same machine as its associated kernel. This offers various
+advantages over the current situation, including:
+
+- Kernels will no longer need to implement the 'heartbeat' for frontends to
+  check that they are still alive.
+- We will be able to interrupt remote kernels (SIGINT cannot be sent over the network)
+- There will be a consistent way to start kernels without a frontend
+  (`jupyter kernel --kernel x`).
+- Kernel stdout & stderr can be captured at the OS level, with real-time updates
+  of output.
```
rgbkrk: :+1:

kernel-nanny.md
takluyver: Thanks, tweaked

takluyver:
Thanks all!

@willingc, yes, it would be good to see your drawing, to check if the explanation conveyed what I was thinking clearly.

willingc:
@takluyver Here's the link to the drawing's folder on Dropbox: https://www.dropbox.com/sh/kzc9bom60c9e57x/AAAWcdlGo8RZB9cklEv7jC2ua?dl=0

takluyver:
Thanks, that looks good.

willingc:
@takluyver Great. You detailed things out very clearly :key:

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
+# Kernel 'nanny' processes
+
+## Summary
+
+We propose to start Jupyter kernels through a 'nanny' process, which will always
+be running on the same machine as its associated kernel. This offers various
+advantages over the current situation, including:
+
+- Kernels will no longer need to implement the 'heartbeat' for frontends to
+  check that they are still alive.
```
n-riesco: How would the nanny process check  the kernel is alive?

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+- We will be able to interrupt remote kernels (SIGINT cannot be sent over the network)
```
n-riesco: How about a message like [shutdown_request](http://jupyter-client.readthedocs.org/en/latest/messaging.html#kernel-shutdown)?

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+- There will be a consistent way to start kernels without a frontend
+  (`jupyter kernel --kernel x`).
```
n-riesco: Perhaps I'm misunderstanding the proposal. Is `jupyter kernel` the nanny? I mean `jupyter kernel` is kernel-agnostic and it'll be able to launch any kernels?

kernel-nanny.md
n-riesco: other potential problems:
- kernels echoing writes to stdout/stderr from the higher level to the lower level
- kernel output cannot be linked to an execution cell, this will cause problems when the user selects the frontend option to run all the cells of a notebook

kernel-nanny.md
n-riesco: Given that the nanny process is going to run in the same machine as the kernel, it makes sense that the nanny process is asked to interrupt the kernel by means of a message similar to `shutdown_request`, then the nanny process interrupts the kernel process by sending the appropriate signal.

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+
+Kernels that opt in to output capturing should print a delimiter (*TODO: define
+delimiter*) on each of stdout and stderr, before and after running user code.
+The delimiter will include the message ID of the execute_request message.
+The nanny will not forward these to the frontend, but will use the 'before'
+delimiters to indicate which execution output resulted from, and the 'after'
+delimiters to detect when stream output is finished. The nanny will tell the
+frontend when all output from an execute_request, on iopub and the two pipes,
+is complete (*TODO: using status:idle, or a new message?*).
+
```
n-riesco: This is assuming synchronous execution.

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+of whether the kernel has enabled output capturing.
+
+### Kernel logging
+
+Where kernels have been previously using low-level stdout/stderr to log to the
+terminal, they need a new way to produce diagnostic logs which shouldn't be
+displayed in the frontend. Kernels opting in to output capturing will be
+started with an environment variable `JUPYTER_KERNEL_LOG` set. The kernel
+should treat this as a filesystem path, which it can open and write logs to.
+
```
n-riesco: Suggestion: add a new type of message for kernels to send their log messages to the frontend.

kernel-nanny.md
takluyver: e.g. `subprocess.Popen.poll()`, but depending on how it's written, there may well be smarter ways. On Unix, the parent process is sent `SIGCHLD` when one of its children dies.

kernel-nanny.md
takluyver: `shutdown_request` is sent to tell the kernel to shut itself down. However, if it's currently executing code, it won't process that message - or any other message - until it finishes. Interrupting is how we break it out of execution.

kernel-nanny.md
takluyver: `jupyter kernel` will be a way to launch the nanny, yes (it may not be called exactly that, but that's the idea).

kernel-nanny.md
takluyver: Right, that's exactly how this will work. We're just trying to work out what form the message will take. If all the world was Unix, we'd almost certainly just call it `signal_request`, and pass a signal number or name. But things get a bit more complicated when we consider kernels running on Windows.

kernel-nanny.md
takluyver: >    kernels echoing writes to stdout/stderr from the higher level to the lower level

Can you expand on this? I don't follow what you mean.

> kernel output cannot be linked to an execution cell, this will cause problems when the user selects the frontend option to run all the cells of a notebook

That shouldn't be a problem: the notebook does that by queueing an execution request for each cell. It has to do that for the output from each cell to go to the right place.

kernel-nanny.md
takluyver: It is, but that's already generally our assumption, and I don't see any other way to do it. There is no side channel in sync with stdout/stderr by which we can convey metadata like the parent message ID, so it has to be in-band.

This is all in addition to the existing mechanisms for kernels to send output, so kernels for which async is really important should focus on capturing output in process and sending the correct messages.

n-riesco:
I would suggest that this proposal is split into four proposals:
1. a proposal to replace SIGINT.
2. a proposal for kernels to capture the low level stderr and stdout streams and forward them to the frontend.
3. a proposal to introduce the command `jupyter kernel --kernel x`
4. a proposal (specific to IPython) that will use a nanny process to implement the first two proposals.

kernel-nanny.md
takluyver: Thanks for the suggestion, but this logging is what you're going to be using if your kernel's messages are not getting to the frontend for whatever reason. So I think it really needs to be a) a separate channel, and b) as technically simple as possible.

takluyver:
> a proposal (specific to IPython) that will use a nanny process to implement the first two proposals.

This is absolutely not specific to IPython - the proposal is for the nanny process to be used for all kernels.

Your 1 & 2 don't really make sense without a nanny process. 3 is doable, but it's a more incidental benefit. I don't see the benefit of splitting this up into smaller pieces: it's one change to the architecture that lets us do a number of useful things, which seems like exactly the right scope for a JEP.

This was also discussed at the in-person dev meeting, and while I don't want to suggest that it's closed for discussion, we did spend some time hashing out what we wanted, and I'd really hope that the remaining issues to work out are details, not the fundamental nature of the proposal.

kernel-nanny.md
n-riesco: For instance, when `console.log('Hello,World!')` was executed, earlier versions of IJavascript would send an iopub message, and would also print 'Hello, World!' to the terminal where the IJavascript was running.

kernel-nanny.md
n-riesco: That's a kernel implementation detail. IJavascript doesn't suffer from that problem (suffers from others, though :stuck_out_tongue: ).

kernel-nanny.md
n-riesco: I would let the kernel handle that. If the kernel can identify the source a stream message, let the kernel send the appropriate iopub reply.

n-riesco:
On 21/04/16 17:54, Thomas Kluyver wrote:
> Your 1 & 2 don't really make sense without a nanny process.

I think the kernel is in a better position to handle 1 and specially 2 than an agnostic nanny process:

- a kernel, if really needed, can implement its own nanny process to handle 1 and 2

- the nanny process cannot determine the origin of stdout and stderr without the kernel's help

- a kernel can always capture the low level stdout/stderr

lbustelo:
@takluyver How about for kernels that are remote? I know this is not officially supported by the notebook server, but it is something that we've experimented with and could be a requirement in certain deployments.

jasongrout:
@n-riesco - the idea in the proposal is that rather than having every kernel implement the capturing and signal/interrupt logic, we'd implement it once outside of the kernel and everyone automatically benefits. As for capturing output, that's opt-in for a kernel, so a kernel absolutely can do their own input/output instead of having the nanny handle it. The nanny makes it much easier to have this automatically taken care of.

jasongrout:
Another concern brought up in the meeting was the latency introduced in forwarding messages through the nanny. Can you mention that in the proposal? I thought @minrk said he might run some tests to get some idea about how much the latency on messages would be impacted by this proposal.

n-riesco:
@takluyver I'm sorry for suggesting a proposal split.

Here are 2 suggestions to the current proposal:
- make the nanny an opt-in feature declared in the kernel spec (currently, IJavascript can run in the official docker images for Node.js; if the nanny was to be made compulsory, then the nanny (and all the dependencies pull by the nanny) would have to be installed in the docker container).
- consider extending the jupyter protocol with a message for the kernel to record log messages (so that frontends have access to them).

jasongrout:
@minrk - how heavy-weight do you see the nanny being? I imagined either a python file, or a lightweight OS-specific C program with zeromq as a dependency.

kernel-nanny.md
takluyver: That's precisely the kind of reason that output capturing will be opt-in - kernels need to be ready for it, and then kernel authors can flip the switch to enable it. It won't be enabled for earlier versions of IJavascript, so that won't be a problem.

kernel-nanny.md
takluyver: Yes, if the kernel is sending the stream message itself, it should absolutely do that. This is for the case where the data is going over the stdout/stderr pipes. If you can capture all stdout/stderr well enough within IJavascript, there's no need for you to enable output capturing.

takluyver:
@lbustelo The idea is that the nanny and the kernel are always running together on the same system. They may both be remote from the frontend (e.g. the notebook server), and this will work much like it already does - zmq messages sent over the network. One of the key advantages of this is that will allow interrupting remote kernels, which is currently impossible.

@n-riesco @jasongrout I definitely see the nanny as being a lightweight thing with few dependencies. In the first instance, it will likely be written in Python, because that's what we can write and debug most effectively, but I may later use it as an excuse to brush up on a language like Rust or Go, which will make it even lighter.

> consider extending the jupyter protocol with a message for the kernel to record log messages (so that frontends have access to them).

The logging system is what you'll want to rely on to debug problems with the messaging, so I want it to be a) a separate mechanism, and b) as simple as possible, like 'open this file and write to it'. We can still arrange things so that the frontend can get the kernel logs by making the log file a named pipe, and having the frontend read it. I like this Unix-y approach here, because it provides a lot of flexibility while requiring very little complexity in the kernels.

n-riesco:
On 22/04/16 14:32, Thomas Kluyver wrote:
> [...] We can still arrange things so that the frontend can get the kernel logs by making the log file a named pipe, and having the frontend read it.

How would that work in the case of remote kernels?

takluyver:
The frontend could ask the nanny to set up a named pipe and forward the data over the network. But let's deal with that when we come to it.

kernel-nanny.md
rgbkrk: From the perspective of writing a native app, we'll do this directly: sending `shutdown_request` first followed by shutting the kernel process down directly.

kernel-nanny.md
rgbkrk: This brings me back to what I was thinking when we talked about the kernel nanny, which is that we should consider a general daemon for launching kernels on a system, similar to docker's interface (CLI + API).

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+object which reads the kernelspec to find how to start the kernel, writes a
+connection file, and launches the kernel process. With the proposed changes, it will
+instead launch the kernel nanny on the machine where the kernel is to run, and
+the nanny will be responsible for creating the connection file and launching
+the kernel process.
+
+**Rejected alternative:** One kernel nanny process per machine, able to start
+multiple kernels. This would be more complex, but we may come back to it later
+if the overhead of one nanny per kernel is too much.
+
```
rgbkrk: Ah, this is down the one nanny per kernel approach now. Glad you cleared this up.

rgbkrk:
One glaring thing that worries from a purely ideological standpoint (and possibly a complexity standpoint): this changes the dynamic of a fairly simple kernel level messaging spec to a "kernel + nanny" level messaging spec.

takluyver:
I don't think it's too big a change ideologically. The main effect is that there will be some extra messages to document between the frontend and the nanny, which the kernel won't directly send/receive.

rgbkrk:
Can frontends still communicate with the kernel without using the nanny? I think that's where @n-riesco is curious (as am I), as the lead of the IJavaScript kernel and a Hydrogen dev.

lbustelo:
Yea, this needs to be optional. Now we are putting a requirement that the system that is hosting the kernels need to support Python.

minrk:
I think KernelNanny should not be optional. Adding it as an option only increases the complexity of everything, because capabilities are now fragmented across use cases.

n-riesco:
Arguably, it'd be simpler not to have a nanny at all, and let kernels handle themselves.

There's however a solution that would make things simple for everyone.

If a kernel nanny was implemented as an optional proxy (that interfaces with frontends in the usual manner), then no changes to the current frontends or kernels (or containers running these kernels) would be required, while it would also let kernels have a nanny if they really needed.

takluyver:
As discussed, the functions of a nanny are not something kernels can readily handle themselves:

- Capturing a process' own std streams at the OS level is fiddly on Unix, and we haven't even worked out how on Windows (if it's possible, it's likely to be something completely different than Unix). Capturing it from the parent process is trivial.
- Interrupting blocked execution: even in languages that do lots of async stuff, unless you can guarantee that nothing will ever block execution, signals need to be sent from another thread or  process.
- Heartbeat: similarly, this should never be blocked by user code executing - though I know that it is in IRkernel, for instance, because I do not want to mess around with threads in R.

The simplest way to achieve all of these things is with a separate process. And I'd rather have one implementation of that that we control and standardise than 50 different ones with their own bugs. Providing this as part of our infrastructure means we're asking less of kernels.

I don't think requiring Python on the server is that big a hardship - most traditional servers will have it anyway, and it's an extra 30MB or so on Docker images, even if it's not already there. However, if it is a problem, we can look into rewriting the nanny in a lower level language.

minrk:
I should clarify what I meant about optional. I didn't mean that it should be impossible to run a kernel without a nanny *ever*, but that it should be a strict requirement for use with `jupyter notebook`.

There are two problems that the nanny solves:

1. remote signaling, restarting, process management
2. (optional) output capture

The first is a need of the application, the second is a need of the kernel. So some *applications* may require there to be a Nanny (e.g. the jupyter notebook application), while some *kernels* may require it (e.g. hypothetical kernels that rely on output capturing). This proposal requires no changes at all to kernels that don't opt-in to output capture, and applications would also have to opt-in to the nanny for the monitoring/signaling support. Hydrogen and nteract would still be able to talk to IJavascript without a nanny, but Jupyter Notebook would not, and IJavascript would not need any awareness that the nanny is present or not.

n-riesco:
On 25/04/16 10:48, Min RK wrote:
> Hydrogen and nteract would still be able to talk to IJavascript without a nanny, but Jupyter Notebook would not, and IJavascript would not need any awareness that the nanny is present or not.

I don't think that's what would happen. Currently, the standard to launch a kernel is its kernel spec.

If a nanny becomes the standard way to launch a kernel, nteract and Hydrogen will be forced to use the nanny, because there would be no other standard to launch the kernel.

takluyver:
This would make one change, as I intend it: by deprecating the heartbeat as something kernels are required to support, applications using a remote kernel will need the nanny to detect kernels dying unexpectedly. Applications could still check for local kernels by monitoring the process. In practice, there are already kernels that don't properly implement the heartbeat (e.g. IRkernel).

@n-riesco the proposal is that the nanny uses the kernelspec, so it will still be there, and other things can still use it.

minrk:
This doesn't change kernelspecs, so that should be unaffected. However, if kernels begin to rely on the nanny features (e.g. new kernels not implementing output capture), you would need the nanny for it to work properly. I would certainly encourage all frontends to use the nanny, but they would not be forced.

n-riesco:
I feel confused. Here are a few questions just to be clear:

1. Will a kernel spec remain the standard to launch a kernel? By this I mean, a frontend will be able to launch a kernel simply by processing its kernel spec (regardless of the use of a nanny).

2. In the case of kernels that make use of a nanny, will frontends be able to invoke the nanny directly? or will they have to make sure 'jupyter kernel' is installed along with the frontend?

minrk:
@n-riesco sorry for the confusion. I believe the answer to both of your questions is yes. The proposal right now is to move the *invocation* of the kernelspec to the nanny process, but not change anything about the spec's `argv` being the command that launches the kernel itself. The change is moving which process invokes that command from the `jupyter notebook` server process to the `jupyter kernel` nanny, which is the process that is started by `jupyter notebook` when a kernel is requested. Related to 2., applications that require a nanny (i.e. `jupyter notebook`) would *always* invoke the nanny directly, trusting the nanny to spawn the kernel; kernels would never invoke the nanny. Applications that prefer to avoid the nanny can choose to only invoke the kernel by way of the nanny if the kernel requires it, but would be able to invoke kernels that don't require a nanny directly from the kernelspec if they so choose.

minrk:
@janschulz I think from Python, if following the conventions of nbconvert and/or notebook, nanny would always be used. There's some ironing out of the Python API to do, though (i.e. does KernelManager stay as the object that instantiates kernels directly and move to the Nanny process, or does KernelManager stay as the application-level API for requesting a new kernel). I do not think it will be a flag to KernelManager, though. I think it will be two different objects for instantiating the nanny and the kernel, only one of which would be used in the application, the other in the nanny itself.

janschulz:
@minrk from the knitpy side, "nanny only" kernels without any influence on how these kernels are started, work fine as long as the MultiKernelManager API to start the kernels is kept :-)

kernel-nanny.md
```diff
@@ -0,0 +1,117 @@
...
+the frontend. Control and iopub (see output capturing) will be connected through
+the nanny, i.e. each channel will have one socket for communications between
+the frontend and the nanny, and a second socket for communications between the
+nanny and the kernel. (*TODO: What are these called in the connection file? Or
+do we have two connection files?*) The heartbeat will only be between the
+frontend and the nanny, to detect situations such as network failures.
+
+## Messaging changes
+
+* A new message type on the control channel from the frontend to the nanny,
```
lbustelo: Why not have the nanny capture https://ipython.org/ipython-doc/3/development/messaging.html#kernel-shutdown

lbustelo:
As it stands, the contract between the kernel and the frontend is a message spec over 0mq. The result is that I can write a kernel and it can be collocated or remote and the kernel does not care (sorry to bring remote kernels again, but I think we need to keep it in mind). It seems to me that this proposal may introduce some complexity to kernels since they not need to be aware of the presence or absence of a nanny. It may also create a split in the kernel landscape between those that want and don't want to be nannied.

This proposal seems to have to parts but too tied together. I think we need to have a:

1. A set of protocol changes that DO NOT mention the word nanny and just specify the content of the messages and desired effect on the kernel side

2. A nanny process that implements some of these management message and can be **optionally** started alongside the kernel.

takluyver:
> sorry to bring remote kernels again

Again, a key part of the reason for this proposal is to facilitate remote kernels, e.g. allowing them to be interrupted. This is not something we're overlooking, it's deliberately part of this.

>  It may also create a split in the kernel landscape between those that want and don't want to be nannied.

No, because that decision is not up to the kernels. They will all work with the nanny, and they should not need to be aware of the presence of the nanny. Applications can choose to use the nanny or include that functionality themselves, but it shouldn't affect how the kernel works: the kernel shouldn't need to know about its parent process.

> This proposal seems to have to parts ... A set of protocol changes .. A nanny process

Again, the messaging changes do not really make sense without an intermediary process. The messaging changes were designed to accommodate the nanny process, not vice versa.

lbustelo:
> Applications can choose to use the nanny or include that functionality themselves, but it shouldn't affect how the kernel works: the kernel shouldn't need to know about its parent process.

@takluyver I totally agree, but I also think that the client should not care if a certain message is handled by some nanny process or whatever the kernel hosting environment chooses to implement. That is why I'm recommending that the protocol piece does not prescribe an architecture or mention the nanny process.

> Again, the messaging changes do not really make sense without an intermediary process. The messaging changes were designed to accommodate the nanny process, not vice versa.

I think the proposal is describing how the local implementation of the proposed messages would be done... a nanny process, etc. and that I'm fine with. But if I want to host kernels using Kubernetes or Mesos, the messaging spec should not have a say on how and what handles these messages.

parente:
I heard part of this discussion at the dev meeting and am catching up here. Two questions (maybe addressed above but eyes missed it):

1. If someone wants to host kernels without a nanny, does this imply that the kernels themselves will need to be updated to support the new messages in a new version of protocol spec when the consumer of those kernels (e.g., notebook server) updates to talk that new protocol version? (Sounds like yes.)

2. Would notebook server become a provider and packager of the kernel nanny for all kernels people want to plug into the notebook? I ask because this has implications on kernel gateway.

takluyver:
1. Whether to use the nanny or not would be an application decision, not a kernel decision. If the application chooses, all kernels will be run through the nanny. If the application doesn't use it, it may need to reimplement some of the nanny features itself (like output capturing, if kernels expect that). No changes are required to kernels - though heartbeats are deprecated, so kernels can stop dealing with that.

2. Yes, installing the notebook server would install the nanny. It would likely be packaged as part of (or a dependency of) `jupyter_client`.

parente:
>  If the application doesn't use it, it may need to reimplement some of the nanny features itself (like output capturing, if kernels expect that). 

Thanks. Now I understand @rgbkrk's question above:

> Can frontends still communicate with the kernel without using the nanny?

If the answer is "no", then is concern about impacting the fledgling ecosystem of new, non-Jupyter notebook clients starting to use kernels (e.g., the ones here: http://blog.ibmjstart.net/2016/03/21/powered-by-jupyter/)? They'll either need to run a nanny or implement whatever support kernels demand from a nanny (e.g., the output capturing you mentioned above). 

Seems like it raises the bar for reusing Jupyter tech pretty significantly.

takluyver:
My expectation is that the default option will be for applications to use the nanny, and this shouldn't be any more difficult for clients than the current setup.

nteract may choose to implement a JS nanny, as it has reimplemented the rest of our Python architecture in JS, but this shouldn't be significantly harder than anything it has already done. As far as I'm aware, the other clients rely on our existing Python stack, so I would expect them to use the nanny. If including the Python runtime is a larger concern, we could also rewrite the nanny as a standalone program in a language like Rust or Go (both of which I'd be happy to have an opportunity to try).

rgbkrk:
You'll have me signed up if the nanny is more like a kernel daemon, which we write in go and distribute as a direct binary.

lbustelo:
So let me back to why I'm keen on understanding this PR. I'm a maintainer of a kernel (Apache Toree). I want to clearly understand what are the implications for me as a kernel developer with regards to this proposal. Here are a list of my concerns:

1. What changes are necessary in the kernel? It is not clear to me from reading all this if I'm going to have to make code changes to support this. There seem to be some contradicting comments above.
2. What are the implications in terms of having the kernel operate under the variety of clients (Jupyter and others) that may or may not opt to use the nanny? Backwards compatibility is VERY IMPORTANT. Not everyone is on the latest and greatest.
3. How does this impact the installation of the kernel in jupyter notebooks?

takluyver:
1. There should be no changes *needed* to the kernel. With the kernel nanny, you *may* stop handling the heartbeat (once you're happy that it won't be a problem for older clients), and you *may* ask the nanny to capture stdout/stderr at the OS level and forward them to the frontend. But it should work with kernels as is.
2. There should be no implications for the kernel. Clients may see some new messages, but in general clients shouldn't crash on unexpected messages.
3. I don't think it should change anything about kernel installation.

takluyver:
@rgbkrk Do you mean having one nanny process per machine, shared between multiple kernels? We had said no to that on the grounds of keeping things simple, but I wonder if the 'daemon nanny' approach, besides having an amazing name, might actually be simpler, because if you assume that the nanny is always running, you can use it to launch remote kernels. This would make it something like a lower-level version of kernel gateway.

rgbkrk:
Yeah, shared between multiple kernels.

My primary motivation is consistency for launching and connecting to kernels, especially amongst the editors using Jupyter kernels (Atom, vim, emacs) as well as Electron based frontends.

toddrjen:
Will clients be able to decide whether to connect to an existing nanny if one exists or always open a new one?  I ask, because my understanding is that it is currently hard to set it up to properly launch jupyter notebooks or qtconsoles from the an application launcher (such as the windows start menu), since each time you do so it will launch a new kernel, leaving many kernels running in the background over time as you open and close the notebook or qtconsole.

takluyver:
That's not entirely accurate: when you close down a Qt console, it does stop the kernel it started. With the notebook, the issue is not (directly) kernels, but the server itself: there's no good way to shut the server down unless it's running in a terminal where you can press Ctrl-C. Separately, there's a difficulty when starting the Qt console from a menu, because the option to connect to an existing kernel is given at the command line, so there's no easy way to use it.

However, none of this really changes with this proposal.

frmdstryr:
The nanny proposal sounds a lot like a controller running on each system. 

Has anyone investigated the feasibility of running all the kernel IO in a thread so requests don't get blocked during execution/apply requests? If this were done, then adding the signal_request message mentioned earlier and a handler that only interrupts if it's not in the IO thread (either use thread.interrupt_main() or  spawn a separate temporary process to interrupt itself) would then work.

Also a bit hackish... but you could redirect sys.stdout sys.stderr to a logger and which tee's the output to the original streams (by saving a reference and redefining). I've done this before in a QT application to route all print statements to a UI logger, just requires some checks to make sure no logging loops are created.

takluyver:
> Has anyone investigated the feasibility of running all the kernel IO in a thread so requests don't get blocked 

Threads are tricky to get right - I'm sure it's doable, but I don't want to encourage every kernel author to do it (and on behalf of IRkernel, I don't want to deal with threads in R). The kernel nanny is something we can implement once and all kernels can benefit.

> you could redirect sys.stdout sys.stderr to a logger and which tee's the output to the original streams

I'm not sure what the advantage of this would be. We already replace `sys.stdout` and `sys.stderr` to route output to the frontend. The issues we're trying to solve here are that:

1. Not all output we want to capture goes through those objects
2. We want a way to produce output that isn't captured, for debugging the kernel itself. Currently, we rely on 1 for this, but if we fix that, we need another way to send output.

kernel-nanny.md
janschulz: For the windows problems, see here: https://github.com/jupyter/jupyter_client/issues/104

janschulz:
One question I've had after rereading the discussion here is how a kernel which is opting in to the new capturing mode would add stuff to the (notebook server) log instead of the notebook UI. Currently the kernel just prints to stdout/stderr and it shows up in my kernel window (Windows cmd box) but not in the UI (just as I want it).

Everything else sounds fine from the perspective of IRkernel: finally a process which can kill IRkernel if it is stuck in a loop :-)

minrk:
@janschulz that's certainly a tricky one, and something we've discussed. When we brought this up earlier, @takluyver noted that it could be solved by creating a named pipe for the process to log to, which the nanny could then dump to its own stderr. I'm not sure what the Windows equivalent would be.

takluyver:
@janschulz this is addressed in the section 'kernel logging' - kernels opting in to output capturing will get a path which they can open and write to. What that path represents would be configured in the frontend - it could just be logging to a regular file, or it could be a named pipe or a pty slave.

It looks like Windows [does have named pipes](http://stackoverflow.com/questions/13319679/createnamedpipe-in-python#13319731), but it's not clear whether you can open and write to them like a regular file or if you need special API calls. If necessary, we can probably do a poor man's pipe with a regular file and the equivalent of `tail -f`.

janschulz:
Windows has named pipes: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365590(v=vs.85).aspx and if I read this right, it seems R also [supports them](https://stat.ethz.ch/R-manual/R-devel/library/base/html/connections.html) as `fifo(description, open = "", blocking = FALSE, encoding = getOption("encoding"))`.

takluyver:
The challenge is whether you can write to an existing named pipe with the regular `open` function, or whether you need to know it's a named pipe. The great thing about the Unix model, and what the 'kernel logging' section here relies on, is that the same APIs (open/write) work on different kinds of filesystem objects. So you can give the kernel a path and it doesn't need to know what it is.

Obviously we could have a second variable that tells the kernel 'this is a fifo', but that leads to extra complexity and probably-untested code in kernels. Worse, it's extra complexity in exactly the bit you want to be simple, because it's the bit you rely on to identify problems. And it's not elegant.

takluyver:
And, to be clear, the question isn't just if we can `open` a pipe with R, but whether in general Windows lets you write to a named pipe without knowing it's a named pipe. If the frontend is going to set up a named pipe, we have to be sure that that it's going to work for whatever kernel it's starting.

janschulz:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365150(v=vs.85).aspx includes this:

> The unique pipe name. This string must have the following form:
> \\.\pipe\pipename
> The pipename part of the name can include any character other than a backslash, including numbers and special characters. The entire pipe name string can be up to 256 characters long. Pipe names are not case sensitive.

-> You can give a pipe a name, but it can't be anything in the regular filesystem.

The [R source for pipe](https://github.com/wch/r-source/blob/8496645d62db4d6d5ee3c7ad67a6fd4a1225e00e/src/main/connections.c#L935) is basically two different implementation for *ix vs windows.

And it's buggy:
```
zz <- fifo("foo-fifo", "w+")
writeLines("abc\n", zz) # does not return :-(
```

Already hang my rsession in RStudio (win) :-(

takluyver:
OK, it sounds like Windows named pipes are not file-like enough that we can rely on kernels being able to write to them. So at least for now, let's stick to regular files on Windows.

n-riesco:
I'm linking to https://github.com/nteract/hydrogen/issues/314 .

If the kernel heartbeat is deprecated, frontends will be unable to distinguish a stalled kernel from execution requests that take very long to complete.

n-riesco:
I think, to avoid issue https://github.com/nteract/hydrogen/issues/314 . The heartbeat message shouldn't be deprecated. Instead, the heartbeat functionality should be offered by the nanny as an opt-in feature.

takluyver:
> If the kernel heartbeat is deprecated, frontends will be unable to distinguish a stalled kernel from execution requests that take very long to complete.

What distinction are you making here? Stalling generally means it's stuck in an infinite loop, but there's no general way to determine if a piece of code will finish or not.

If the kernel crashes, the nanny can detect this without requiring heartbeat messages - that's one of the key reasons for this proposal.

rgbkrk:
> Stalling generally means it's stuck in an infinite loop, but there's no general way to determine if a piece of code will finish or not.

You're telling me we can't solve the halting problem as part of the project? Sheesh.

takluyver:
Well, we have to limit our scope somehow, and after some contentious discussion, we decided that things which have actually been proved impossible should be out of scope for another year or two. ;-)

n-riesco:
By busy, for example, I mean a kernel running an execution request like this:
```python
a = 1
while True:
    a += 1
```

By stalled, for example, I mean a kernel that has run the following:
```python
import os
os.system("kill -STOP " + str(os.getpid()))
```

takluyver:
Oh, you mean something has actually suspended the process. That doesn't seem like it would come up very often. If we need it, it looks like there are ways to check if a process is suspended, though they may not be the same across platforms.

I specifically want to get rid of heartbeats because not all kernels can easily implement it properly (e.g. IRkernel won't respond to heartbeat pings while it's processing other messages). We also don't use heartbeats in the notebook, so we know things work OK without them.

takluyver:
I'd like to move this forwards, because we've been promising people something with these capabilities for a long time.

I think the main remaining thing to discuss is whether we use one nanny per kernel (so the frontend starts a nanny process for each kernel it wants) or one nanny per system (so the frontend asks the nanny to start kernels). We had considered this at the dev meeting and opted for nanny-per-kernel, at least for now, as it seems considerably simpler. However, a nanny-per-system approach could provide an easy way to start kernels on a remote system. Then again, that assumes that you want multiple kernels per system; if you're instantiating containers to run one kernel each, having a multi-kernel-capable nanny in each one is unnecessary complexity.

So I still lean towards nanny-per-kernel for now, but I'd try to design the architecture so that it could use one nanny for multiple kernels in the future.

AlexTugarev:
@takluyver, after catching up with the discussion here I tend to speak up for the nanny-per-kernel strategy. The idea of decoupling to make kernels remotely interruptable is great and it also justifies the higher architectural complexity. Though the later should be kept as low as possible. 

We are developing a kernel that requires a process which cannot handle interrupt signals as intended by jupyter and the kernel would be terminated on interrupts. Therefore we already have a surrogate process in place which translates an interrupt signal into a kernel interrupt notification. 

Two questions: 
1. What would be the difference between nanny-per-kernel and nanny-per-system in terms of interrupting the kernel?
2. How it is planed to start the nanny on the machine where the kernel is to run?

Regarding the output synchronization: in (edge) case of background threads printing to std streams the kernel is not able to detect the end, but may decide to reply. I'd suggest to capture only output between the delimiters and ignore the rest, cf. https://github.com/jupyter/jupyterlab/issues/629.

takluyver:
Thanks Alex,

> 1. What would be the difference between nanny-per-kernel and nanny-per-system in terms of interrupting the kernel?

There should be no difference; in both cases, the frontend will send a message to the nanny asking it to generate an interrupt signal for the kernel.

I haven't thought about kernels that may not work with signals. What kind of notification do you use instead of a signal to interrupt the kernel?

> 2. How it is planed to start the nanny on the machine where the kernel is to run?

The default case will still be local kernels, in which case the nanny will run as a subprocess of the frontend. For remote machines, it depends on which model we go for: in nanny-per-kernel, the frontend will use whatever means it currently has to start a kernel to start a nanny instead, and the nanny would start the kernel. For nanny-per-system, the nanny would probably be configured as a service, so it would already be running, and the frontend would connect to it to ask it to start a kernel.

AlexTugarev:
> I haven't thought about kernels that may not work with signals. What kind of notification do you use instead of a signal to interrupt the kernel?

Handling of signals in a Jvm process is no fun, because SIGINT terminates the process. That is why we introduced a surrogate process which is started first and is restarted on SIGINT while the actual kernel keeps running. 

If you are about to touch the protocol, we would suggest to introduce a simple 'interrupt_request' message to be sent to the kernel. And it's not just about Java discourages custom signal handlers, we see a disruption to the meta-level in interrupting the kernels via signals while using a high-level protocol for everything else.  Would it make sense to abstract the kernel from the process? The control message which proposed in this PR would fit the concept. The kernel nanny could ask the kernel for its capabilities and could still do the signaling for compatibility reasons. 

With the introduction of 'interrupt_request' messages, we would disable the built-in Jvm signal handlers and deal with the request message internally.

takluyver:
We use signals specifically because they can interrupt execution; without that, you have to have something awkward with threads or separate processes just to receive the message (as you do), and then you still need some way to interrupt the kernel execution (which I'm still not sure how you do - what does the surrogate process do to the kernel to stop whatever code it's executing?).

I'm open to adding a different interrupt mechanism that kernels can ask the nanny to use, but I'm still pretty puzzled by this:

1. I'm surprised there's no way to handle SIGINT on a JVM. So any time you Ctrl-C a Java program, it has no choice but to die?
2. I don't understand what you do instead to interrupt execution. I understand that there's a surrogate process to receive the signal on behalf of the kernel, but I don't see how it then interrupts the kernel.

AlexTugarev:
> ... you have to have something awkward with threads or separate processes just to receive the message (as you do), and then you still need some way to interrupt the kernel execution ...

I isn't that awkward to work with threads in Java, it actually can be very efficient. In our case the engine of the kernel is running as a separate thread and that engine has an API to interrupt the execution.

> 1. I'm surprised there's no way to handle SIGINT on a JVM. So any time you Ctrl-C a Java program, it has no choice but to die?

The default Jvm signal handler will terminate the process, therefore the simple answer is Yes. There is actually a discouraged internal API to provide a custom `sun.misc.SignalHandler`, which might be removed in the future.

> 2. I don't understand what you do instead to interrupt execution. I understand that there's a surrogate process to receive the signal on behalf of the kernel, but I don't see how it then interrupts the kernel.

When the process is about to terminate on SIGINT we can execute some code. What we do now is to create a marker file, but its about to be proper IPC. Within the actual kernel process we are able to detect that the proxy process was restarted because of an interrupt signal.

> I'm open to adding a different interrupt mechanism that kernels can ask the nanny to use ...

That sounds great.

takluyver:
> In our case the engine of the kernel is running as a separate thread and that engine has an API to interrupt the execution.

Does that work even if it's in a blocking syscall? It sounds... suboptimal.

Anyway, if we do add another mechanism for interrupting kernels, what makes sense? I'd be inclined to go for something simple like a named pipe that you can do a blocking `read()` from, and write a byte for an interrupt, but that might be tricky on Windows. It could be yet another ZMQ socket.

rgbkrk:
Bump from the dev meeting. We should allow some clearance to iterate on this with code, beyond that of this proposal.

takluyver:
From discussion with Min today, we're leaning towards nanny-per-system (multiple kernels per nanny). I'm hoping to find time to work on this again.

jasongrout:
> From discussion with Min today, we're leaning towards nanny-per-system (multiple kernels per nanny). I'm hoping to find time to work on this again.
>  👍 1

Why the switch from nanny-per-kernel? I'm glad you might work on this soon!

takluyver:
There are advantages either way; this way allows the nanny to be running as a service, so remote frontends have a standard way to connect, see which kernel specs are available, and start one. Neither of us felt strongly, but I want to pick something so I can go and prototype it.

Even with nanny-per-system, I'd still like to provide an entry point to start the nanny and immediately start one kernel, for the cases where you're spinning up a VM or container to run a single kernel.

jasongrout:
> I want to pick something so I can go and prototype it.

Definitely +1!

lbustelo:
When thinking about remote frontends, it starts to feel like this may be something that belongs in kernel gateway. See https://github.com/jupyter/kernel_gateway_demos/tree/master/nb2kg.

takluyver:
The proposed 'kernel nanny', now that we're leaning towards one nanny managing multiple kernels, is indeed quite like kernel gateway. The key difference is that this would expose a ZeroMQ interface, rather than an HTTP/websockets interface.

1. Do we still want to do this, or should we always use HTTP for managing remote kernels? I see some potential advantages of exposing a ZMQ interface, but I don't know if they're compelling enough:
  - Security could be simpler in some situations as it doesn't rely on SSL certificates.
  - There may be more (or better supported) zmq libraries for some languages than websocket client libraries, as the main purpose of websockets is to talk to a browser. I haven't surveyed zmq & websocket client libraries, though.
2. If we do want a ZMQ interface to managing kernels, would it share much code with kernel gateway? Would it make sense to be part of the kernel gateway project?

tristanz:
As a consumer, I find ZMQ and custom encryption to be a cost compared to WebSockets + SSL.

takluyver:
Right, there are definitely situations where that's easier. But I could imagine other situations where dealing with SSL certificates is more complex than a simple shared secret for the connection.

lbustelo:
@takluyver Checkout the suppor in kernel gateway for personalities. Might be that it can get enhanced to expose ZMQ.

parente:
The kernel gateway is using the notebook server code as a package and so only knows how to communicate with the outside world using HTTP/Websocket based protocols. Adding a zmq personality would be quite an undertaking.

Also, I'm not sure what the future holds for KG in general. Notebook has subsumed some of its features (e.g., token auth, kernel activity monitoring) and jupyter_server is on the enhancement proposal table for separating the frontend and backend bits of notebook.

minrk:
websockets definitely have some nice characteristics when talking over remote networks (dealing with all of the connections over a single port is a big one). I think supporting additional transports for remote kernels is a bit of a separate discussion, though. Adding the KernelNanny should make such a proposal easier to implement, since we would know that we always have a process running next to the kernel which could serve as that multiplexer. Part of the point of the nanny proposal is that it is transparent to both the kernel and the client.

The most basic functionality that I want from KernelNanny is the original idea in [IPEP 12](https://github.com/ipython/ipython/wiki/IPEP-12:-Splitting-KernelManager-and-KernelClient#never-a-kernel-without-a-kernelmanager):

- KernelClient gets the interrupt/restart/etc methods of KernelManager that can be expected to work for all kernels.

Since the kernel client API is all zmq, it makes the most sense to me for these to be zmq request/reply messages like all the rest, rather than adding additional HTTP requests to an otherwise zmq API. If we went with HTTP, these would presumably be regular HTTP requests, though, not websockets, so the bar is lower for clients to talk to them. It does seem internally inconsistent to have some zmq requests and some HTTP requests for the same API, though.

To me, starting remote kernels, which is solved by KG, is not part of the nanny proposal. But they do creep closer together when we make the nanny a singleton kernel-providing service rather than a simple manager of one kernel, since it has to gain start/list APIs in addition to the more basic stop/restart.

takluyver:
I was roughly thinking that with local kernels, the server would integrate the nanny functionality (i.e. it would keep handling interrupts as it currently does), and the remote case would be handled by nb2kg (or a ZMQ equivalent if we wrote one). I suspect it would be no harder to write a KernelManager/KernelClient pair wrapping the HTTP/websockets interface than to write a ZMQ kernel gateway.

Maybe this points back to the nanny being per kernel, if we've already got KG as the multiple kernel manager.

I think I need to think more carefully about what situations we actually want remote kernels in, and how we manage them (e.g. putting kernels in docker lends itself to one kernel per container; does the nanny process need to be inside the container with the kernel? Or can it do what it needs from outside?).

minrk:
> I was roughly thinking that with local kernels, the server would integrate the nanny functionality (i.e. it would keep handling interrupts as it currently does)

I think that makes sense. The QtConsole and jupyter_console and all other entrypoints would also need to run the nanny or talk to an existing one.

> putting kernels in docker lends itself to one kernel per container; does the nanny process need to be inside the container with the kernel? Or can it do what it needs from outside?

I think it can go either way. The simplest version is the nanny in the container with the kernel, where there's really no awareness that docker is involved. The more sophisticated version is a "DockerKernelNanny", akin to the DockerSpawner in JupyterHub, where 'raw' kernels are in containers and the nanny does all of its activities via the docker API.

kernel-nanny.md
rolweber: Can the kernel log output be disabled? If yes, how is that indicated? Variable not set, or set to an empty value?
As it is described now, the nanny would have to create an empty file and set JUPYTER_KERNEL_LOG for every kernel that uses output capturing, even if that kernel does not produce diagnostic logs. Disabling log output from the kernelspec, for example, would be a way to avoid that overhead.

thomasjm:
Hi all -- sorry to comment on an old thread, but I'm hoping to get clarification about something...what is the current status of the deprecation of kernel heartbeats? There are some mentions of deprecating the heartbeat in this proposal, but the latest version of the [messaging spec](http://jupyter-client.readthedocs.io/en/latest/messaging.html) that I can find still contains them.

This is troubling me because I just noticed a kernel ([gophernotes](https://github.com/gopherdata/gophernotes)) which seems to have dropped heartbeat support in July 2017. I somehow missed the memo about this whole thing, so could someone let me know if heartbeats are dead and the nanny process is how things work now? Is there another source of documentation/truth besides http://jupyter-client.readthedocs.io ? Thanks!

takluyver:
No problem!

The Jupyter notebook doesn't rely on heartbeats, and since that's the interface most people use, some kernels only worry about supporting it. We could make all interfaces work without the heartbeat so long as the kernel runs on the same machine (which it normally does). I don't remember which ones already work like this.

Getting rid of the heartbeat entirely was waiting on this proposal because it would have provided another way to tell if a *remote* kernel is still alive.

thomasjm:
Got it, thanks! Is there some official way to keep up-to-date on the latest version of the messaging spec? Speaking as someone working on an alternate Jupyter frontend, dropping heartbeats seems like a major change that would merit a new major version number on the "Messaging in Jupyter" page...

takluyver:
If we ever do officially drop them, that would certainly be a new protocol version. For now, kernels are theoretically expected to implement it, but in typical cases there's no need, so many don't. We should probably clarify that in the messaging doc.

thomasjm:
Cool, thanks for the clarification. Tagging @dwhitena so he sees these comments.

dwhitena:
Thanks @thomasjm. We will work on supporting this in gophernotes.

sushantd195:
> a proposal for kernels to capture the low level stderr and stdout streams and forward them to the frontend.
This is still an issue as seen here. I'm on latest IRkernel and reticulate
image

cladosphaero:
Hi @takluyver @jasongrout was this ever implemented? It looks like OS-level output is still suppressed by Jupyter, as evidenced by reticulate not displaying Python stdout in IRkernel.

krassowski:
There is an adjacent pre-proposal which addresses some of the same issues (but not IRkernel logging issue) over at https://github.com/jupyter/enhancement-proposals/issues/117.

Zsailer:
Hi all 👋 —Zach here from the @jupyter/software-steering-council.

PR #16, opened 2016-06-10
Localizing Notebooks - continued

srl295:
Continuation of: https://github.com/jupyter/enhancement-proposals/pull/10

Based on @TwistedHardware ’s commits. (I dropped the ones which modified the base template.)


jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,118 @@
...
+<a href="#">New Notebook</a>
+```
+
+This will be done like this:
+
+```HTML
+<a href="#">{{ _("New Notebook") }}</a>
+```
+
+### Javascript files
```
takluyver: I don't think this section is accurate - it looks like it's left over from when the proposal was to turn all JS files into templates, which we agreed we don't want to do.

JCEmmons:
Updated proposal for translation of Jupyter notebook.  Please review/approve, as I'm able to work on this now that I'm back on the job.

rgbkrk:
I'm certainly in favor of pressing forward. Let's see how it plays out.

parente:
I agree with Kyle. The best way to vet the approach is to try it out in a few PRs against notebook and adjust accordingly.

As for the proposal document, I took a quick read through. I want to go back and make a few minor suggestions.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,214 @@
...
+At install time, convert the translated `.po` into two runtime formats:
+* Convert to `.mo` which can be used by Python code using the gettext() APIs in Python, and can also be used by
+the i18n extensions in [Jinja2](http://jinja.pocoo.org/docs/dev/extensions/#i18n-extension)
+
+* Convert to JSON using [po2json](https://github.com/mikeedwards/po2json) for consumption by the Javascript
+code within Jupyter.
+
+## Detailed Explanation
+
+The Jupyter notebook code presents a significant challenge in terms of enablement for translation,
```
parente: A link to https://github.com/jupyter/notebook over the text "Jupyter notebook code" might help clarify that the document is explicitly talking about the classic notebook UI and not Jupyter Lab. Given the differences between the two implementations, I think it wise to focus on the former (if that is indeed your goal) and, based on lessons learned from that, tackle Lab localization as a separate task.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,214 @@
...
+mostly because there are multiple different types of source code from which translatable UI strings
+are derived.
+
+In Jupyter, translatable strings can come from one of three places:
```
parente: "In Jupyter Notebook"

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,214 @@
...
+        env = Environment(loader=FileSystemLoader(template_path), extensions=['jinja2.ext.i18n'], **jenv_opt)
+        env.install_gettext_translations(trans, newstyle=False)
+```
+
+Note here that **trans** is the same variable initialized by `gettext.translation()` in the Python example earlier.
+
+### Javascript files
+
+For Javascript, there are no established APIs that can consume a compiled `.mo` file directly in the same way as gettext().
+However, there is a library called [Jed](https://slexaxton.github.io/Jed/) that provides an API set similar to gettext().
```
parente: Is Jed compatible with webpack and/or requirejs (both of which are used in Jupyter Notebook)?

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,214 @@
...
+[Babel](http://babel.pocoo.org/en/latest/),
+[Jed](https://slexaxton.github.io/Jed/),
+[po2json](https://github.com/mikeedwards/po2json).  There may be some difficulties with the licensing.
+* Jupyter notebook developers have to be made aware of the proper ways to externalize any new strings
+that are added, and to perform message extraction via [Babel](http://babel.pocoo.org/en/latest/)
+whenever strings change.
+
+
+## Prototype - Proof of Concept
+I have created a **VERY** preliminary prototype of Jupyter notebook at (https://github.com/JCEmmons/notebook/tree/intl)
```
parente: Oh, awesome! 🍰

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
JCEmmons: I'm no javascript guru, but I think I would be, since the example on its [main page](https://slexaxton.github.io/Jed/) shows using it as an AMD module.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
JCEmmons: Good idea, Peter. I added these links and a scope section that makes it clear that this proposal is only for classic notebook.  See 6c00c531b5bfbdeb7354ccfcb277461492996aec

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,214 @@
...
+```
+at the top of the python code and then adding
+```python
+# Set up message catalog access
+trans = gettext.translation('notebook', localedir=os.path.join(base_dir, 'locale'), fallback=True)
+trans.install()
+```
+
+Once this is complete, any calls to `_()` in the code will retrieve a translated string from 
+${base_dir}/locale/**xx**/LC_MESSAGES/notebook.mo, where **xx** is the language code in use
```
JCEmmons: Done. See bff04d5b39de40c208b477c56dc31c02fa81d8cd

minrk:
Picking this up after the holiday. The only thing I have left is that this proposal is scoped only the "notebook classic" as opposed to JupyterLab, and I think the reverse makes a bit more sense to me. JupyterLab is the future of the notebook client, and the notebook package is heading for retirement pretty soon. Plus, the discussion of tooling sounds like it will fit better in the existing JupyterLab pipeline than the notebook classic one. What do people think about that?

minrk:
My brain is rusty, I misunderstood the tooling question, because I read [babel](http://babel.pocoo.org/en/latest/) as [babel](https://babeljs.io). So that part doesn't seem relevant to notebook vs JupyterLab. Given that the server code is still the `notebook` package for both sides, I think this is a sensible proposal to start there.

Approved by me as-is.

JCEmmons:
Thanks, and while I understand that Jupyter lab is the future, "classic" is the present at least for us.  I'm close to being ready for a preliminary PR on this, so I will welcome your comments and advice.

fperez:
As long as @JCEmmons doesn't mind the fact that there may be a bit of work to be re-done for JLab, I'm +1 here.  In fact, there may be some advantage in doing it in Classic so we can kick the tires of the process atop a fairly stable and simpler codebase. Lab is more complex and moving rapidly, so I suspect doing this in Lab right now could prove quite tricky.

@JCEmmons, since you seem OK with that, I'd say let's go ahead and see what we learn!

@minrk pinged the steering council (thanks!) for feedback, but we'll treat silence as acquiescence if we don't hear back from the rest in a few days.  No need to stall you here, and very sorry for the delay.  We really appreciate  your contribution, and this is an important problem we've long failed to tackle.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,224 @@
+# Jupyter Notebook Translation and Localization
+
+## Problem
+
+There is currently no standard approach for translating the GUI of [Jupyter notebook]( https://github.com/jupyter/notebook).
+This has driven some people to do a
+[single language translation for Jupyter 4.1](https://twitter.com/Mbussonn/status/685870031247400960).
```
takluyver: I may be wrong, but I think @Carreau's tweet was pointing to a translation of the release announcement, not the UI.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+Once this is complete, any calls to `_()` in the code will retrieve a translated string from 
+${base_dir}/locale/**xx**/LC_MESSAGES/notebook.mo, where **xx** is the language code in use
+when the notebook is launched.
+For example "de" for German, "fr" for French, etc. If no message catalog is available, or if
+the string doesn't exist in the catalog, then the string passed as the argument to _() is
+returned.
+
+In this context, **${base_dir}** refers to the base installation directory for notebook. We are
```
takluyver: This definition is not entirely clear to me - I can think of a couple of directories that may be called the 'base installation directory'. I'm guessing you mean the notebook package directory (which contains, for instance, `notebookapp.py`).

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,224 @@
...
+### Javascript files
+
+For Javascript, there are no established APIs that can consume a compiled `.mo` file directly in the same way as gettext().
+However, there is a library called [Jed](https://slexaxton.github.io/Jed/) that provides an API set similar to gettext().
+[Jed](https://slexaxton.github.io/Jed/) uses JSON as it's input file instead of `.mo` files, but
+the good news here is that there are plenty of 3rd party utilities that can convert from `.po` into a form of JSON that
+can be consumed by Jed. The author of [Jed](https://slexaxton.github.io/Jed/) recommends
+[po2json](https://www.npmjs.com/package/po2json), which can be used
+either as a command line utility, or directly from Javascript.  I suspect that the conversion from `.po` to either
+`.mo` for Python or Jinja2, or conversion to JSON for Javascript would be something we would want to do at install time.
```
takluyver: Probably at build time to be precise - i.e. at least the wheel packages will contain pre-built `.mo` and `.json` translation files.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,224 @@
...
+
+Identification of strings for translation using this method is done similarly to the way we would do it for Python:
+by creating a function named `_()`, enclosing all translatable strings as an argument to this function, and then
+binding it to the `gettext()` API.  The binding in Javascript would look like this:
+
+```javascript
+    var i18n = new Jed(nbjson);
+    var _ = function (text) {
+    	return i18n.gettext(text);
+    }
```
takluyver: Does this have to be done before any JS code uses translatable strings? If all UI construction has to wait for an asynchronous request to get the JSON file, I can see that causing problems. Maybe we can stick the JSON data into the page when we fill the template to avoid that.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,224 @@
...
+        .attr('title', _('Click here to rename, delete, etc.'))
+        .appendTo(item);
+}
+```
+
+## Use of Babel to extract translatable strings
+
+[Babel](http://babel.pocoo.org/en/latest/) allows us to define a set of rules that define
+where all the extractable strings are (whether Python, HTML template, or Javascript), what the
+extraction methods are (i.e. `_()` for Python or Javascript, and `<% trans >`/`<% endtrans %>`
```
takluyver: `<>` -> `{}`, based on the Jinja section above

JCEmmons:
Great - you can view the work in progress on this item at 
https://github.com/JCEmmons/notebook/tree/jep16.  I am hoping to have a 
preliminary PR ready soon.


Regards,

John C. Emmons
Globalization Architect & Unicode CLDR TC Chairman
IBM Software Group
Internet: emmo@us.ibm.com

takluyver:
I spotted a few minor things that I think should be fixed (see above). Once those are dealt with, I think we can merge this.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
Carreau: Indeed. I often translate at least the release announce in French.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,224 @@
...
+[Jupyter notebook]( https://github.com/jupyter/notebook).
+
+## Proposed Enhancement
+
+Use [Babel](http://babel.pocoo.org/en/latest/)
+to extract translatable strings from the Jupyter code, creating `.pot` files that can be updated
+whenever the code base changes. The `.pot` file can be thought of as the source from which all translations are derived.
+
+Translators and/or interested contributors can then use utilities such as [Poedit](https://poedit.net/) to create
+translated `.po` files from the master `.pot` file for the desired languages.
```
Carreau: We did experiment with https://www.transifex.com/ipython/ipython-notebook/ which is online, and we hoped at some point to integrate this into the workflow.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,224 @@
...
+
+At install time, convert the translated `.po` into two runtime formats:
+* Convert to `.mo` which can be used by Python code using the gettext() APIs in Python, and can also be used by
+the i18n extensions in [Jinja2](http://jinja.pocoo.org/docs/dev/extensions/#i18n-extension)
```
Carreau: what do you mean by "install", does this require code execution ? If so we should find an alternative way, as wheels cannot execute code at install time.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Detailed Explanation
+
+The [Jupyter notebook code]( https://github.com/jupyter/notebook) presents a significant challenge in terms of enablement for translation,
+mostly because there are multiple different types of source code from which translatable UI strings
+are derived.
+
+In [Jupyter notebook]( https://github.com/jupyter/notebook), translatable strings can come from one of three places:
+
+1. Directly from Python code
```
Carreau: Nitpicking,  and not really important to address here, but we might need to clarify that this will not/cannot address translations of message coming from the kernel. IE `?`/`??` will likely not be translated.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
Carreau: I think allowing to ship translation as external package would be good. Otherwise releases might get hold on by missing translations. I think trying to pull that from `notebook-localisation` (or similar), or even `notebook-i18n-lang` would be good for end-user translator to be able to work without having to install dev with node and all this stuff. 
But that can be fixed later. I'm happy to leave that as-is, and modify this after the fact if we figure out how to implement it.

jupyter-notebook-gui-translation/jupyter-notebook-gui-translation.md
Carreau: I think with caching it should be ok, as we can query the file by its hash and make it expire in 99years. It depends on the size though.

Carreau:
Sorry for the delay. Minor comments none of them blocking this. 
I'm pretty sure we can (and should) allow translation to be installed separately. I have no issue allowing some translation to be shipped with the notebook, but I believe _if_ external translation are installed, we should pick these. 

One other point about translating string I haven't seen is the use of `{identifier}` instead of `%s` (because translation may have a different order), but that's minor.

takluyver:
There is a related proposal for bi-directional support in the notebook:
https://github.com/jupyter/notebook/issues/2178

PR #23, opened 2017-09-20
Add Template as Metatada enhancement proposal

rgaiacs:


takluyver:
@mpacer was working on machinery to allow custom nbconvert behaviour from the notebook web server, in https://github.com/jupyter/notebook/pull/2413, so I imagine he'll be interested in this.

parente:
If I'm reading the proposal properly, I think it's possible to implement some of the proposed use cases via [bundler extensions](http://jupyter-notebook.readthedocs.io/en/stable/extending/bundler_extensions.html) in the notebook server. But bundlers do not necessarily capture format information in notebook documents. They only provide a way for extension writers to plug in arbitrary, server-side actions to take on notebook documents (e.g., running nbconvert on a notebook with a specific template, posting a notebook to some external API).

akhmerov:
I'm wondering about possible security and privacy implications of this proposal. Could one use an online template to track users of the notebook and/or provide malicious code? Could one use a path to a private local file to get access to it?

rgaiacs:
> I'm wondering about possible security and privacy implications of this proposal.

Thanks to raise the questions.

> Could one use an online template to track users of the notebook and/or provide malicious code?

Probably yes. Someone could use, for example, [Google Analytics](https://analytics.google.com/) or [Facebook (Social) Page Plugin](https://developers.facebook.com/docs/plugins/page-plugin/) to track users. But as far as I know, nbconvert doesn't do anything to block someone to track users with the tools that I just mention.

Malicious code is a much bigger issue. HTML template should be quite safe due the "sandbox environment" created by the browsers. Markdown template and RST template should also be safe since they are only plain text files that can not be executed. Unfortunately, PDF, Python, R, Julia and other programming language template could have malicious code, for example, they could have instructions to remove files recursively from the user home directory or from the file system root.

I agree that we should avoid the possibility of someone include malicious code on the template but from my point of view end-users should be as much careful when using the template or the file generate by the template as they are when using any open source project. Every time that I go to GitLab, GitHub, Bitbucket or another source code repository and I download one project to test there is a small change that the code that I just download has some malicious code. I just need to trust other users.

> Could one use a path to a private local file to get access to it?

[Is possible to add functions to Jinja2 to do it](https://stackoverflow.com/a/9769454/1802726) but not without passing extra information to Jinja2 of which file it should load the new functions.

akhmerov:
>  But as far as I know, nbconvert doesn't do anything to block someone to track users with the tools that I just mention.

AFAIK if you download an untrusted notebook and nbconvert it, the action of nbconverting is impossible to track.

> HTML template should be quite safe due the "sandbox environment" created by the browsers.

Can you explain what you mean here? Imagine we're dealing with a notebook that the user converts to upload on their company blog, and it starts leaking user account information since it has all domain permissions. Or am I missing something here?

rgaiacs:
> AFAIK if you download an untrusted notebook and nbconvert it, the action of nbconverting is impossible to track.

But you can track who downloaded the untrusted notebook. GitHub, GitLab, Bitbucket, Google Drive, Dropbox, ... probably do some kind of tracking.

I didn't mentioned to my proposal because I though it would be kind of obvious, but users will also have the option to change/remove the default custom template without losing any relevant information, i.e. the content of the cells. Any user that is concern to be being track by the server who host the custom template could just remove the custom template to use the default one.

> > HTML template should be quite safe due the "sandbox environment" created by the browsers.
>
> Can you explain what you mean here?

W3C specification says that web browsers **must** ask users to access any local file. This is for security reasons to avoid someone to start reading your files when you visit one web page. And browsers should also ask permission from users when load content from third party servers, i.e. if you are accessing `x.com` your browsers should avoid load content from `y.com`. Also, browsers should manager cookies and any other storage in a way to preserve users privacy.

> Imagine we're dealing with a notebook that the user converts to upload on their company blog, and it starts leaking user account information since it has all domain permissions. Or am I missing something here?

# Static web site

Although Jinja2 is powerful and this can create security issues, is hard to me how a good staff will convert a Jupyter Notebook blog post to be published on his company blog using his company Jupyter Notebook template and leak user account info. I can only imagine that the staff went rogue.

# Dynamic web site

I can imagine why some companies could want to use a template to save staff time by converting Jupyter Notebook into Python/R/Julia/... code that is going to be run in production for a website. A custom template will for sure create some security issues since one staff could went rogue and use a different template. I would say that if the company is accepting this new code with any review or without force using a trusted custom template for the conversion on their continuous integration deploy, see my next sentence, they have serious issue. As I mentioned at https://github.com/jupyter/enhancement-proposals/pull/23/files#diff-10de2d1c00eb5cbe8087028c71d82f6bR66, when user run

~~~
$ jupyter nbconvert --to language  --template=custom_template.tpl 'example.ipynb'
~~~

nbcovert should use `custom_template.tpl` first event if the Jupyter Notebook says that another custom template should be use.

akhmerov:
> Although Jinja2 is powerful and this can create security issues, is hard to me how a good staff will convert a Jupyter Notebook blog post to be published on his company blog using his company Jupyter Notebook template and leak user account info. I can only imagine that the staff went rogue.

A likely scenario is that this person downloaded an interesting notebook from the web, then modified it, changing the topic completely, then forgot about the source.

A different questions: the use cases that the JEP describes both refer to the uses where intended granularity level is notebook server and not single notebooks. Would a notebook server extension or a new configuration option not be more a appropriate solution?

Zsailer:
Hi @rgaiacs 👋—Zach from the @jupyter/software-steering-council here.

PR #28, opened 2019-02-12
Jupyter-server

Zsailer:
After chatting with @SylvainCorlay, I'm sharing a new draft of the Jupyter Server enhancement proposal. This supersedes #21. 

Check it out! Any feedback is welcome.

You can easily view a rich display version of the proposal [here](https://github.com/jupyter/enhancement-proposals/pull/28/files?short_path=389bcd1#diff-389bcd1b95a4cf4726133a8728985375).

For those interested, I've been experimenting with some of the ideas in this proposal. Check out these repos (warning: all are **experimental**!):

* [Zsailer/notebook_server_extension](https://github.com/Zsailer/notebook_server_extension): The classic Jupyter Notebook as a server extension.
* [Zsailer/jupyter_server_extension](https://github.com/Zsailer/jupyter_server_extension): The easy way to write Jupyter server extensions that launch as Jupyter applications.

Pingin: @SylvainCorlay, @blink1073, @kevin-bates, @lresende, @ellisonbg, @minrk, @bollwyvl

### DIFF
--- a/jupyter-server/jupyter-server.md
+++ b/jupyter-server/jupyter-server.md
@@ -1,5 +1,15 @@
 # Standalone Jupyter server enhancement proposal [active]
 
+| Item       | Value                                                                                                                        |
+|------------|------------------------------------------------------------------------------------------------------------------------------|
+| JEP Number | 10                                                                                                                           |
+| Title      | Standalone Jupyter Server                                                                                                    |
+| Authors    | Zach Sailer ([@Zsailer](https://github.com/Zsailer)) and Sylvain Corlay ([@SylvainCorlay](https://github.com/sylvaincorlay)) |
+| Status     | Draft                                                                                                                        |
+| Type       | S - [Standards Track](https://www.python.org/dev/peps/#pep-types-key) JEP                                                    |
+| Created    | 20 September 2016                                                                                                            |
+| History    | 20 September 2016, 12 February 2019                                                                                          |
+
 ## Problem
 
 There are now multiple frontends that talk to the backend services provided by the notebook server: the legacy Notebook, the dashboards, JupyterLab, standalone widgets and more. The configuration of legacy notebook and the backend server are tightly coupled. As a consequence, the other applications are forced to load the legacy notebook to use the backend server.
--- a/jupyter-server/jupyter-server.md
+++ b/jupyter-server/jupyter-server.md
@@ -37,10 +37,7 @@ Decouple the backend server (and its configuration) from the classic notebook fr
 1. Namespacing static files and REST API urls. 
     - Each extension serves static files at the `/static/<extension>` url.
     - New `ExtensionHandlerApp` class
-1. Migrate configuration.
-    - Notebook App configuration stays in `jupyter_notebook_config.py`
-    - Server and services configurations move to `jupyter_server_config.py`
-    - Add a `migrate` application to automate this migration.
+1. Migrate application.
 1. Add necessary documentation to notebook and jupyter_server repos.
 
 ## Detailed Explanation
--- a/jupyter-server/jupyter-server.md
+++ b/jupyter-server/jupyter-server.md
@@ -18,27 +18,27 @@ There are now multiple frontends that talk to the backend services provided by t
 
 Decouple the backend server (and its configuration) from the classic notebook frontend. This will require the following steps:
 
-1. Separate `jupyter_server` repository
+1. [Separate `jupyter_server` repository](#separate-jupyter_server-repository)
     - A fork of the `notebook` repository. 
     - Pure Python package with notebook backend services.
     - `notebook` tornado handlers and frontend logic stay in `notebook` repo.
     - Deprecated notebook server APIs do not move to `jupyter_server`.
-1. Clearly define the smallest unit that is a "Jupyter Server"
+1. [Clearly define the smallest unit that is a "Jupyter Server"](#clearly-define-the-smallest-unit-that-is-a-jupyter-server)
     - Kernel, KernelSpecs, and Session (?) become the only core services
     - Everything else becomes a server extension?
-1. Extensions as Applications
+1. [Extensions as Applications](#extensions-as-applications)
     - server extensions move to `jupyter_server`.
     - New `ExtensionApp` class
     - Notebook, jupyterlab, etc. become ExtensionApps. 
-1. New server extensions mechanism.
+1. [New server extensions mechanism.](#new-server-extensions-mechanism)
     - new base classes to create applications from server extensions. 
     - nbextensions stay in `notebook`. 
     - `jupyter_notebook_config.d` folder becomes `jupyter_server_config.d`
-1. Namespacing static files and REST API urls. 
+1. [Namespacing static files and REST API urls.](#add-namespacing-to-static-endpoints-and-rest-api-urls) 
     - Each extension serves static files at the `/static/<extension>` url.
     - New `ExtensionHandlerApp` class
-1. Migrate application.
-1. Add necessary documentation to notebook and jupyter_server repos.
+1. [Configuration System](#configuration-system)
+1. [Migrate application.](#migration-application)
 
 ## Detailed Explanation
 
@@ -167,7 +167,7 @@ Here are some steps we can take to reduce the friction for transitioning:
 
 For an thorough overview of the configuration system and a side-by-side diff between the (current) notebook and (future) server config layout, take a look at [Zsailer/jupyter_config_overview](https://github.com/Zsailer/jupyter_config_overview).
 
-**Migration application**
+### Migration application
 
 To make migration easier on users, a `migrate` application will be available to split notebook and server configurations into their appropriate locations (listed above).
 
--- a/jupyter-server/jupyter-server.md
+++ b/jupyter-server/jupyter-server.md
@@ -1,0 +1,216 @@
+# Standalone Jupyter server enhancement proposal [active]
+
+## Problem
+
+There are now multiple frontends that talk to the backend services provided by the notebook server: the legacy Notebook, the dashboards, JupyterLab, standalone widgets and more. The configuration of legacy notebook and the backend server are tightly coupled. As a consequence, the other applications are forced to load the legacy notebook to use the backend server.
+
+## Proposed Enhancement
+
+Decouple the backend server (and its configuration) from the classic notebook frontend. This will require the following steps:
+
+1. Create `jupyter_server` repository
+    - Fork of `notebook` repository. 
+    - Pure Python package with notebook backend services.
+    - `notebook` tornado handlers and frontend logic stay in `notebook` repo.
+    - Deprecated notebook server APIs do not move to `jupyter_server`.
+2. New server extensions mechanism.
+    - server extensions move to `jupyter_server`.
+    - new base classes to create applications from server extensions. 
+    - nbextensions stay in `notebook`. 
+    - `jupyter_notebook_config.d` folder becomes `jupyter_server_config.d`
+    - `notebook` becomes a server extension.
+3. Services become server extensions
+    - Services are just serve extensions with dependencies.
+    - Add dependency injection system.
+5. Namespacing static files and REST API urls. 
+    - Each extension serves static files at the `/static/<extension>` url.
+    - 
+6. Migrate configuration.
+    - Notebook App configuration stays in `jupyter_notebook_config.py`
+    - Server and services configurations move to `jupyter_server_config.py`
+    - Add a `migrate` application to automate this migration.
+7. Add necessary documentation to notebook and jupyter_server repos.
+
+## Detailed Explanation
+
+### Create a `jupyter_server` repository
+
+The first thing to do is fork `notebook`. A new `jupyter_server` repo will keep the server specific logic and remove:
+1. the notebook frontend code, 
+2. deprecated notebook server APIs, and
+3. tornado handlers to the classic notebook interface. These pieces will stay in the `notebook` repository.
+
+Things that stay in notebook:
+
+- `edit` module: the handlers for the classic notebook text editor.
+- `templates` directory: all html templates for the classic notebook
+- `terminal` module: handlers for the classic notebook terminal application.
+- `view` module: handlers for file viewer component.
+- `static` directory: all js and style sheets for notebook frontend.
+- `tree` module: a classic notebook file browser+handlers.
+- `auth` module? *(Should this move to jupyter_server?)*
+
+Things that move to jupyter_server:
+- `services` module: all jupyter server services, managers, and handlers
+- `bundler`: handlers for building download bundles
+- `files`: handlers for serving files from contents manager.
+- `kernelspec`: handler for getting kernelspec
+- `base`: base handler for jupyter apps.
+- `i18n`: module for internationalizing the jupyter server
+
+
+Preliminary work resides in [jupyter_server](https://github.com/jupyter/jupyter_server).
+
+### Server Extensions
+
+The extension mechanism for the *jupyter server* will be the main area where server extensions differ from notebook's server extensions.
+
+Enabled server extensions will still be loaded using the `load_jupyter_server_extension` approach when the jupyter server is started. 
+
+**Server extensions as applications**
+
+In the proposed jupyter_server, extension developers may also create an application from their extension, using a new `JupyterServerExtensionApp` class. Extension developers can subclass the new base class to make server extensions into Jupyter applications (configurable and launchable from the commmand line). This new class loads extension config from file and parses configuration set from the command line. 
+
+For example, the legacy notebook could be started: 1) as an enabled extension or 2) by running the traitlets application from the command line.
+
+Example extension:
+```python
+from .extension import load_jupyter_server_extension
+
+class NotebookApp(JupyterServerExtensionApp):
+
+    name = 'notebook'
+    description = 'NotebookApp as a server extension.'
+    load_jupyter_server_extension = staticmethod(load_jupyter_server_extension)
+```
+
+`JupyterServerExtensionApp` subclasses are configurable using Jupyter's configuration system. Users can generate config files for each extension in the Jupyter config path (see `jupyter --paths`). Each extension's configuration is loaded when the server is initialized or the extension application is launched.
+
+As will all jupyter applications, users can autogenerate a configuration file for their extension using `jupyter my_extension --generate-config`.
+
+Initial experimental work resides in [`jupyter_server_extension`](https://github.com/Zsailer/jupyter_server_extension).
+
+**Classic notebook server extension**
+
+The classic `NotebookApp` will become a server extension. It will inherit `JupyterServerExtensionApp`. Users can enable the notebook using the extension install/enable mechanism or enable the `notebook` in their `jupyter_server_config.py` file:
+```python
+c.ServerApp.jpserver_extensions = {
+    'notebook': True
+}
+```
+Users can also launch the notebook application using the (usual)`jupyter notebook` command line interface. 
+
+**Extension installing/enabling mechanism**
+
+The new extension mechanism in the *jupyter server* will differ from notebook's server extensions.
+
+* The `--sys-prefix` installation would become the default. (Users are confused when enabling an extension requires more permissions than the installation of the package). Installation in system-wide directories would be done with the `--system` option. 
+* Installing an extension will include the addition of a 'manifest' file into a conf.d directory (under one of the Jupyter configuration directories, `user / sys-prefix / system`). The placement of such an extension manifest provided by a Python package can be done with `jupyter server extension install --py packagename [--user / --system / --sys-prefix]`. Packages (conda or wheels) carrying server extensions could place such manifests in the sys-prefix path by default effectively installing them. Development installations would also require the call to the installation command.
+
+* Enabling an extension is separate from the installation. Multiple scenarios are possible:
+  - enabling an extension at the same level (user / sys-prefix / system) as where it was installed, or at a higher level (user for sys-prefix and system, or sys-prefix for system).
+  - forcibly disabling an extension that was enabled at a lower level of precedence.
+  - forcibly enabling an extension that was disabled at a lower level of precedence.
+  This would be done via two `conf.d` configuration directories managing a list of disabled extensions and list of enabled extensions in the form of empty files having the name of the corresponding extension. If an extension is both disabled and enabled at the same level of precedence, disabling has precedence. Packages (conda or wheels) could place such a enabler file in the sys-prefix path by default. The `jupyter server extension enable` command would be required for development installations.
+  
+(Possibly) when an extension is enabled at a given precedence level, it may only look for the version of the extension installed at the same or lower precedence level. For example, if an extension `foobar` is installed and enabled system wide, but a user installs a version with `--user`, this version will only be picked up if the user also enables it with `--user`.
+
+### Services become server extensions (with dependency injection)
+
+Right now, there are two ways to extend and customize the jupyter server: services and server extensions. This is a bit confusing for new contributors. The main difference is that services often (but not always) **depend on other services**. 
+
+On example is the `sessions` service, which depends on the `kernels` and `kernelspec` services. Without these two services, the `sessions` service doesn't work (or even make sense). 
+
+We could reduce complexity around `jupyter_server` by making *everything* a server extension. We would need to add a **dependency injection** system to allow extensions to depend on other extensions. Some good options are [pyinject](https://github.com/google/pinject) or [python-dependency-injector](https://github.com/ets-labs/python-dependency-injector).
+
+To port a service, it will need to be refactored using extension mechanism mentioned above.
+
+### Add namespacing to `static` endpoints and REST API urls.
+
+Currently, the notebook tornado application serves all static files underneath the `/static/` prefix. Jupyter server will add namespacing under the static url and extension REST API urls. Each extension will serve their static files under the `/static/<extension-name>` prefix and their API handlers behind a `/extension/api/<extension-name>` prefix.
+
+For example, the classic notebook server extension will add static handlers that reroute requests to the `/static/notebook/` endpoints.
+
+The jupyter_server will provide a new `JupyterExtensionHandler` base class that reroute requests to the extension's namespaced static and REST API endpoints.
+
+Preliminary experimental work resides in the [`jupyter_server_extension`](https://github.com/Zsailer/jupyter_server_extension) repository.
+
+### Configuration System
+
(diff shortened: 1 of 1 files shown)


willingc:
@rgbkrk @captainsafia @betatim fyi

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+  
+(Possibly) when an extension is enabled at a given precedence level, it may only look for the version of the extension installed at the same or lower precedence level. For example, if an extension `foobar` is installed and enabled system wide, but a user installs a version with `--user`, this version will only be picked up if the user also enables it with `--user`.
+
+### Services become server extensions (with dependency injection)
+
+Right now, there are two ways to extend and customize the jupyter server: services and server extensions. This is a bit confusing for new contributors. The main difference is that services often (but not always) **depend on other services**. 
+
+On example is the `sessions` service, which depends on the `kernels` and `kernelspec` services. Without these two services, the `sessions` service doesn't work (or even make sense).
```
rgbkrk: Core services should stick around, otherwise I think we're going to end up further in this diaspora of endpoints with no concrete aim of what makes up the core of jupyter.

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+    - `notebook` becomes a server extension.
+3. Services become server extensions
+    - Services are just serve extensions with dependencies.
+    - Add dependency injection system.
+5. Namespacing static files and REST API urls. 
+    - Each extension serves static files at the `/static/<extension>` url.
+    - 
+6. Migrate configuration.
+    - Notebook App configuration stays in `jupyter_notebook_config.py`
+    - Server and services configurations move to `jupyter_server_config.py`
```
rgbkrk: Does it have to go in `jupyter_server_config.py` or can we use the general `jupyter_config.py`?

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+
+* Allow various frontends to use the backend services of the jupyter server.
+* Provides base classes for extension developers writing server extension applications and handlers.
+* Reduce complexity when extending the server by making everything a server extension.
+* Organizes the configuration for server and extension in a sane and logical manner (`conf.d` approach).
+
+**Cons** associated with this implementation include:
+
+* Break the classic notebook in a backwards incompatible way.
+* Affects many projects. The transition may be painful?
```
captainsafia: Are there any transition difficulties that we can plan for?

Zsailer:
Thank you @rgbkrk and @captainsafia for looking at this document already! And thank you @willingc for pinging them.

jupyter-server/jupyter-server.md
mpacer: >Does it have to go in `jupyter_server_config.py` or can we use the general `jupyter_config.py`?
We always can so long as it is a standard JupyterApp also respect the `jupyter_config.py`, so I imagine this one would as well

However, I don't think that's the right problem. I think what we should do is make a change to JupyterApp from `jupyter_core` that allows us to declare additional configuration file names that should be picked up. Currently you can only address one app at time (or all apps with `jupyter_config.py`) based on the value of the app's class attribute `name`.

Additionally, we could make a Configurable class (NotebookAppConfig) that inherits from the NotebookApp and overwrites all of it's webserver capabilities so that we can continue to get access to the existing `NotebookApp.nbserver_extension` information. But again… that will be meaningless if we can't read in people's jupyter_notebook_config.py, jupyter_notebook_config.json, and jupyter_notebook_config.d/extension.json.

rgbkrk: Adding on -- with Netflix's current deployments we deploy classic, jupyterlab, and nteract -- we want all of them and in many cases want to share config (so that `ContentsManager`s are the same and additional server extensions are available for APIs, like with [`bookstore`](https://github.com/nteract/bookstore)).

Even if I was elsewhere, I'd still want to have a consistent deterministic config setup even with multiple jupyter apps attached to the server.

Zsailer:
@mpacer Thank you for the review and comments! My day is a little crazy today, but I'll try to get to these points by tomorrow.

jupyter-server/jupyter-server.md
Zsailer: >I think we're going to end up further in this diaspora of endpoints with no concrete aim of what makes up the core of jupyter.

That's a great point. Thanks, Kyle. 

My original motivation for exploring the idea of merging services and extensions came from projects like [kernel_gateway]() and [enterprise_gateway](). These projects only provide kernels as a service and drop the other services. My thought was that we could make "which services are provided" a configurable option. 

By doing this though, I recognize that I'm approaching a point where the jupyter server is just a configurable tornado server. Perhaps this is the wrong move. 

Maybe we need to define what the core jupyter server is--and maybe that's what we have already (minus the notebook). Or maybe "kernels as a service" is the smallest "jupyter unit", and other services should be extensions?

jupyter-server/jupyter-server.md
Zsailer: That's a great question. 

I think we should start listing specific transition difficulties and how to move foward here. I was *kind of* getting at that in the "How this effects other projects" section, but it needs to be fleshed out a lot more.

jupyter-server/jupyter-server.md
rgbkrk: I'd think `kernels` and `kernelspecs` at the very least, as well as `sessions`.

I understand not having `contents` be part of this core service, though maybe that makes `sessions` non-functional(?).

jupyter-server/jupyter-server.md
Zsailer: Thanks @rgbkrk and @mpacer for your comments here. 

>make a change to JupyterApp from `jupyter_core` that allows us to declare additional configuration file names

@mpacer this is a good idea--add a trait to the `JupyterApp` for listing configuration files that should get loaded (and in what order of precedence). That would certainly help with avoiding breaking everyone's configuration system in this transition. 

I missed [jupyter/notebook#4376](https://github.com/jupyter/notebook/pull/4376) when trying to research all the threads around configuration. Sorry about that! I'll need to think about this a bit more for a new draft. (I'm working on adding flow diagram/visualizations showing the proposed configuration models to this proposal to help others follow the conversation as well).

That said, I know that jupyterlab has moved towards a configuration model where each extension has its own config file a `conf.d`. I was modeling the jupyter server (and its extensions) after this model. But, this conversation has made me realize this needs a lot more thought. Specifically, I need to think a lot more about Netflix's use-case here... thanks for sharing @rgbkrk

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+    - nbextensions stay in `notebook`. 
+    - `jupyter_notebook_config.d` folder becomes `jupyter_server_config.d`
+    - `notebook` becomes a server extension.
+3. Services become server extensions
+    - Services are just serve extensions with dependencies.
+    - Add dependency injection system.
+5. Namespacing static files and REST API urls. 
+    - Each extension serves static files at the `/static/<extension>` url.
+    - 
+6. Migrate configuration.
```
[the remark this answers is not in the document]
Zsailer: Thanks @mpacer. 

I think you're absolutely right. This is an **early** draft still, so there should be many more iterations to improve this JEP.

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+    - Notebook App configuration stays in `jupyter_notebook_config.py`
+    - Server and services configurations move to `jupyter_server_config.py`
+    - Add a `migrate` application to automate this migration.
```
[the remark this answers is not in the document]
Zsailer: That's a fair critique. First and foremost, it sounds like I need to rethink the configuration model and flesh out the details. Perhaps we can reach a solution without requiring a migration.

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+7. Add necessary documentation to notebook and jupyter_server repos.
```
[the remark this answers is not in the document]
Zsailer: 👍

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+
+The extension mechanism for the *jupyter server* will be the main area where server extensions differ from notebook's server extensions.
+
+Enabled server extensions will still be loaded using the `load_jupyter_server_extension` approach when the jupyter server is started. 
+
+**Server extensions as applications**
+
+In the proposed jupyter_server, extension developers may also create an application from their extension, using a new `JupyterServerExtensionApp` class. Extension developers can subclass the new base class to make server extensions into Jupyter applications (configurable and launchable from the commmand line). This new class loads extension config from file and parses configuration set from the command line. 
+
+For example, the legacy notebook could be started: 1) as an enabled extension or 2) by running the traitlets application from the command line.
```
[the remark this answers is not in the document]
Zsailer: Sorry, the language is a bit sloppy. 

The idea here is to make server extensions that can *behave* like stand alone applications. You can load them by launching jupyter server and listing them as an extension *or* you can call `jupyter <my_extension>` from the command line (and pass configurable traits to that extension at the CLI). 

The example I mention is a "classic notebook" extension, which launches the classic notebook frontend. You would be able run the notebook application by launching jupyter server and loading the notebook extension *or* calling the extension directly from the CLI (i.e. `jupyter classic-notebook` or maybe `jupyter notebook`).
 
This would be made possible by a `JupyterServerExtensionApp` class (which inherit `JupyterApp`). An extension that inherit this class behaves like a `JupyterApp` that starts a jupyter server and appends the extension to the server. 

As a prototype, checkout out [jupyter_server_extension](https://github.com/Zsailer/jupyter_server_extension).

Does that make sense?

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+**New proposed configuration structure**
+
+The jupyter_server configuration lives in the `jupyter_server_config.py` file. Extensios are configured in separate files in the `jupyter_server_config.d` folder. The notebook configuration is stored in a `notebook.py` file, just like other extensions.
+```
+~/.jupyter/
+├── jupyter_server_config.py
+└── jupyter_server_config.d
+  ├── notebook.py|json
```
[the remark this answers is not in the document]
bollwyvl: hear, hear!

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+from .extension import load_jupyter_server_extension
+
+class NotebookApp(JupyterServerExtensionApp):
+
+    name = 'notebook'
+    description = 'NotebookApp as a server extension.'
+    load_jupyter_server_extension = staticmethod(load_jupyter_server_extension)
+```
+
+`JupyterServerExtensionApp` subclasses are configurable using Jupyter's configuration system. Users can generate config files for each extension in the Jupyter config path (see `jupyter --paths`). Each extension's configuration is loaded when the server is initialized or the extension application is launched.
```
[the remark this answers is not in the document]
Zsailer: >Would this be a proper `JupyterApp` as well or only a `Configurable`?

It would be a `JupyterApp`.

>In which repo would the `JupyterServerExtensionApp` live?

I'm not sure quite yet. Maybe inside the `jupyter_server` repo? Or it could live as a small side repo.

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+
+Things that stay in notebook:
+
+- `edit` module: the handlers for the classic notebook text editor.
+- `templates` directory: all html templates for the classic notebook
+- `terminal` module: handlers for the classic notebook terminal application.
+- `view` module: handlers for file viewer component.
+- `static` directory: all js and style sheets for notebook frontend.
+- `tree` module: a classic notebook file browser+handlers.
+- `auth` module? *(Should this move to jupyter_server?)*
```
bollwyvl: I feel like `auth` should probably be the new server, but _potentially_ have a more pluggable design. Security should be pretty hard-nosed by default, e.g. base handlers that already have `@web.authenticated` on, etc. with it being an explicit decision for some new handler to _not_ be authenticated.

jupyter-server/jupyter-server.md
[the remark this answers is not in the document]
bollwyvl: This name is java-like in length... perhaps `jupyter_server.ExtensionApp`?

jupyter-server/jupyter-server.md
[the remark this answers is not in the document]
Zsailer: Can you point me to this discussion? There are many threads, so it's easy to get lost.

I'm coming to jupyter server conversation/topic *very* late in the game... so I'm working hard to get up to speed, but I'm definitely going to miss key conversations. Its helping me a ton to compile all that information here.

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+├── jupyter_server_config.py
+└── jupyter_server_config.d
+  ├── notebook.py|json
+  ├── lab.py|json
+  └── my_extension.py|json
+```
+
+**Migration application**
+
+To make migration easier on users, the jupyter server will include a `migrate` application to split notebook and server configurations into their appropriate locations (listed above). This application will find any `jupyter_notebook_config.py|json` files in `jupyter --paths`, read configured traits, sort server vs. notebook traits, and write them to the appropriate `jupyter_server_config.py|json` and `jupyter_notebook_config.py|json` files.
```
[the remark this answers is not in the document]
Zsailer: That's a fair point. Perhaps this will go away after I rethink this proposal some more.

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+Users can also launch the notebook application using the (usual)`jupyter notebook` command line interface. 
+
+**Extension installing/enabling mechanism**
+
+The new extension mechanism in the *jupyter server* will differ from notebook's server extensions.
+
+* The `--sys-prefix` installation would become the default. (Users are confused when enabling an extension requires more permissions than the installation of the package). Installation in system-wide directories would be done with the `--system` option. 
+* Installing an extension will include the addition of a 'manifest' file into a conf.d directory (under one of the Jupyter configuration directories, `user / sys-prefix / system`). The placement of such an extension manifest provided by a Python package can be done with `jupyter server extension install --py packagename [--user / --system / --sys-prefix]`. Packages (conda or wheels) carrying server extensions could place such manifests in the sys-prefix path by default effectively installing them. Development installations would also require the call to the installation command.
+
+* Enabling an extension is separate from the installation. Multiple scenarios are possible:
```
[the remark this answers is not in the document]
bollwyvl: Indeed, `*_config.d` was a big step forward, making it far easier to use standard tooling (`pip`, `conda`). As the trend continues more towards ease-of-installation, I'd really not like to see us go back to a two-step process.

Zsailer:
@mpacer Thanks a ton for you comments and feedback! This was extremely helpful.

Zsailer:
In full disclosure, I'm entering this JEP way late in the game. I realize there have been many conversations (and likely debates) in the past that are before my time at Jupyter, and thus, I'm completely unaware of. Please don't hold it against me! 😃 I promise I'm not intentionally ignoring anyone's previous thoughts or comments. I am trying to synthesize everything I come across into this single document. Any links to conversations you know we had in the past would help me tremendously. 

I'm hoping that this PR becomes a great opportunity to gather those opinions/thoughts again in one place and reignite the conversation. I'll do my best to iterate on this document quickly and improve it based on the feedback. I'm just hoping to make progress on this proposal since it's been sitting for a while.  

I posted this before the dev meeting to get some eyes on the documents. Then, maybe it would be a good idea to discuss this deeper in person. Either way, keep sharing your feedback! I really appreciate it!

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Services become server extensions (with dependency injection)
+
+Right now, there are two ways to extend and customize the jupyter server: services and server extensions. This is a bit confusing for new contributors. The main difference is that services often (but not always) **depend on other services**. 
+
+On example is the `sessions` service, which depends on the `kernels` and `kernelspec` services. Without these two services, the `sessions` service doesn't work (or even make sense). 
+
+We could reduce complexity around `jupyter_server` by making *everything* a server extension. We would need to add a **dependency injection** system to allow extensions to depend on other extensions. Some good options are [pyinject](https://github.com/google/pinject) or [python-dependency-injector](https://github.com/ets-labs/python-dependency-injector).
```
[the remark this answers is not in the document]
bollwyvl: I think the intent is for an extension to depend on _some_ implementation of a feature provided by a serverextension that you _might not know the name of_, so can't look for and/or import? I suppose you could walk all of the serverextensions and see if they have the function you want... 

serverextension config issues aside (which will be good to fix, good show!) I think the closest thing we have today are the various `*_manager`s, but first-party code is littered with use of private methods of specific implementations, which extensions see and replicate. However, I don't know if we should be looking to find _more_ dependencies to solve these problems: I'd love to have this server end up with _fewer_. `entry_points` of course come to mind, but with `conf.d` we've mostly solved the discoverability need, but maybe we just need more/different named extension points with tighter contracts than _here's the whole `app`, don't break anything!_

However, at the same time, a lot of the by-inheritance or replacement patterns we have make it hard to compose behaviors. For example, I would like a `ContentsManager` that offered a route scheme, such that I could have, say, `/contents/local/index.ipynb` and `/contents/github/jupyter/notebook/index.ipynb`, that the frontend and other services would transparently work with, but it's either git or local unless I want to write my own ecosystem. Kernelspecs are much the same way: `nb_conda_kernels` still confuses the hell out of people, but is actually a pretty sane way to work on multiple notebook-driven projects, but can't be easily composed with, say, a docker provider and remote provider. @takluyver has made significant strides towards this with the [KernelFinder](https://github.com/jupyter/jupyter_client/blob/44980c13680f4e4226cf25f199ce4e4bb6e11296/jupyter_client/discovery.py#L87) pattern, but we'd need lots more: `ContentsFinder`, `AuthFinder`, `SessionFinder`.

jupyter-server/jupyter-server.md
kevin-bates: @Zsailer - thank you for writing this proposal and starting this great discussion!

@rgbkrk - I totally agree with the notion that kernel management (which needs to include kernelspecs or providers, etc. as you noted) are the root of what "a jupyter server" provides.  It's not clear to me what `sessions` actually provide from a server perspective since clients of the gateways don't need them to spawn kernels.  

Although I'm biased, I believe the ability to optionally ([and easily](https://github.com/jupyter/notebook/pull/4161)) route the kernel management to another node is crucial for multi-tenant capabilities and achieves a natural separation of user data from computation.  I can't think of good reasons to necessarily separate out other pieces but that could be due to my lack of experience outside of the gateways.  I'm not saying that jupyter_server should solely consist of kernel management - not at all.  Just that the "server" should be able to optionally remote certain services.  One service that makes sense for this is kernel management (and by association kernelspec management).

@lresende fyi

jupyter-server/jupyter-server.md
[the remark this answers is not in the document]
rgbkrk: Not sure where the thread is, but I'd expect `jupyter_*_.d` to not be dynamic config, only static json.

bollwyvl:
Here's a thought experiment: what would it take to rebuild nbviewer on top
of the proposed architecture? There's no better free way to get a bunch of
people to test our stuff!

- config-disable all the kernels/spec stuff
- configure a MemCachingProxyContentsManager
- add TheGithubContentsFinder, TheInternetContentsFinder, etc.
- configure a MemCachingNbconvertManager
- add some custom static assets
- add a home to /
- add a route to remap /nbconvert/.* to /.*

Sounds pretty good to me!

This also suggests to me that more things should be async/awaitable, like
contents_manager.get, calling nbconvert.from_notebook_node (as exposed in
the server, not nbconvert itself!)

Zsailer:
Okay, I realized that I didn't fully understand Jupyter's configuration system before. 

So, I took the time to create this ["Jupyter Configuration Overview"](https://github.com/Zsailer/jupyter_config_overview) repo. It shows an example **notebook server** config setup, with READMEs in each subdirectory explaining the various pieces. It also shows an example **jupyter server** config setup for comparison. 

The main README also has a section listing the differences between the two. I tried to make it simple and transparent. It also lists some practical steps to reduce friction for such as a transition. 

I'd love to get your thoughts from those of you who know Jupyter's core much better than me: @mpacer, @rgbkrk, @bollwyvl, and @willingc. I've opened an issue [here](https://github.com/Zsailer/jupyter_config_overview/issues/1) with a summary for discussion. Once that discussion rounds off, I'll fix/update the JEP with our conclusions.

Does that work for others?

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
+# Standalone Jupyter server enhancement proposal [active]
+
+## Problem
+
+There are now multiple frontends that talk to the backend services provided by the notebook server: the legacy Notebook, the dashboards, JupyterLab, standalone widgets and more. The configuration of legacy notebook and the backend server are tightly coupled. As a consequence, the other applications are forced to load the legacy notebook to use the backend server.
```
willingc: Let's add nteract into list of services.

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Proposed Enhancement
+
+Decouple the backend server (and its configuration) from the classic notebook frontend. This will require the following steps:
+
+1. Create `jupyter_server` repository
```
willingc: Hasn't this already been done? https://github.com/jupyter/jupyter_server Or will there be a new_jupyter_server repo?

willingc: Perhaps instead of steps, provide an overview of key functional components that would be stored in a repo together:
- backend server which handles services
- base server extensions provided for access to server information and processes
- services, built upon base extensions, which can inject dependencies
- REST API and namespacing of static files
- configuration for server, base server extensions, and services

willingc: Each functional component would then be broken out as its own section to better scope discussion, options, reject alternatives etc.

Zsailer: >Hasn't this already been done?

👍 Yes, I *am* referring to the (already created) jupyter_server repo. This repo was created after the original draft of this proposal (~2 years ago). "Create" was a poor choice of words. It's already been created. They point is to explicitly state that `jupyter_server` is a product of this proposal.

>Perhaps instead of steps, provide an overview of key functional components

👍 That sounds like a good idea. I had organized each item in my mind as a chronological set of steps to implementation, but there is not practical hierarchy to the pieces in this proposal. For the most part, they are separate pieces that with separate discussion.

Zsailer:
> Overall, I would recommend giving this document a way to be iterated upon. Something similar to [#29 (comment)](https://github.com/jupyter/enhancement-proposals/pull/29#issuecomment-467724022) would apply here too. While the status is DRAFT for the JEP, it would be a work in process and not accepted.

That's a great idea. Thanks @willingc. I'll adjust the proposal to reflect a "draft" status and make it easier to iterate.

jupyter-server/jupyter-server.md
SylvainCorlay: I am not sure if it makes sense to provide tool for the transition, which may end up being harder to use and maintain than doing the change manually.

jupyter-server/jupyter-server.md
```diff
@@ -0,0 +1,216 @@
...
+
+* Enabling an extension is separate from the installation. Multiple scenarios are possible:
+  - enabling an extension at the same level (user / sys-prefix / system) as where it was installed, or at a higher level (user for sys-prefix and system, or sys-prefix for system).
+  - forcibly disabling an extension that was enabled at a lower level of precedence.
+  - forcibly enabling an extension that was disabled at a lower level of precedence.
+  This would be done via two `conf.d` configuration directories managing a list of disabled extensions and list of enabled extensions in the form of empty files having the name of the corresponding extension. If an extension is both disabled and enabled at the same level of precedence, disabling has precedence. Packages (conda or wheels) could place such a enabler file in the sys-prefix path by default. The `jupyter server extension enable` command would be required for development installations.
```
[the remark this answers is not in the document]
SylvainCorlay: These conf.d configuration files would live in `PREFIX/etc/jupyter/server/`.

@mpacer what do you mean about allowing server extension packages to automatically disable other server extensions?

jupyter-server/jupyter-server.md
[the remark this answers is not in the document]
SylvainCorlay: > Not sure where the thread is, but I'd expect `jupyter_*_.d` to _not_ be dynamic config, only static json.

I would agree with that

jupyter-server/jupyter-server.md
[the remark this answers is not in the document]
SylvainCorlay: I am am also doubtful about an auto-migration tool. It might be better to merely document the migration.

willingc:
@Zsailer Perhaps let's do this... 

- [ ] Add a header block similar to what I suggested in the comment and similar to a Python PEP. Use this original PR number as the JEP number. Status is DRAFT.
- [ ] Incorporate any straightforward change from all these comments
- [ ] Larger changes, such as my request to break into sections instead of steps, leave as is for the moment. Let's open individual issues on the repo capturing each discussion point.
- [ ] After the issues are opened, let's merge this PR.

I think then we'll be in good shape to iterate and refine the JEP through smaller subsequent PRs.

Zsailer:
Per @willingc's great suggestion, I've done the following:

- [X] Added a header block marking this as a "Draft"
- [X] Revised the proposal, incorporating some of the changes discussed here (plus some updates from things I've learned along the way).
- [X] Broken each point into separate Issues on this repo (prefixed with [JEP 10]). 
- [ ] Merge PR and iterate later.

Check out the newest draft. We can now iterate on this document through later PRs. And as always, thank you ahead of time for any feedback!

jupyter-server/jupyter-server.md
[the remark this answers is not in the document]
Zsailer: Thanks everyone. I've adjusted this in the new draft of the proposal.

willingc:
I'm going to go ahead and merge this PR after all this great work by @Zsailer. I hope that the process of iterating on the JEP and tackling the issues that were opened will help move forward the direction.

SylvainCorlay:
Many thanks to @Zsailer for pushing this!

rgbkrk:
Thank you for marking it as a draft and getting it moving in a positive direction @Zsailer + @willingc. 😄

Zsailer:
Thank you all for the review! 

I'll work on moving the comment threads left here to the issues I opened for each section in the proposal. I look forward to further discussion. 😄

PR #38, opened 2019-03-04
Adding a distribution section to the JEP JEP and small tweaks

choldgraf:
Hey all - this PR builds off of #29 to add some language to the JEP enhancement proposal. There's a relevant issue to discuss this here: #27 

This PR is primarily about adding a "distribution" section to the "meta JEP". It mentions using a public website as an "archive" of finished JEPs. I tried not to tie it in to any specific technology, since I feel like that is more of an implementation detail and we can cross that bridge when we come to it. Hope that's OK.

The PR also:

* Cleans up a few formatting things I found in there, and adds links to the ipython/python *EP process
* Adds this JEP to the list of active JEPs in the readme (I am a bit unclear that status this JEP has...so I just left what was in the metadata for this one)

Lemme know what folks think!

PR #39, opened 2019-03-06
Minor corrections to spelling, grammar

parente:
Some of these might already be corrected in #38.


choldgraf:
I'm happy to rebase! I'd merge but don't have permissions to do so!

parente:
Ah sorry. Thought ya’ll had permissions. On that note, do we have a plan for who should have ownership on this repo to merge and do general admin? Maybe it’s part of the meta JEP and I’m not remembering. If not, should we call that out explicitly?

PR #43, opened 2019-10-21
add JEP 42 (voila incorporation) to readme

Zsailer:
I realized we forgot to have you add JEP #42 to the front page README.

@SylvainCorlay


SylvainCorlay:
@blink1073 you beat me to it.

SylvainCorlay:
I am so glad I got number 42.

choldgraf:
question for @SylvainCorlay - is this JEP now "implemented"? If not, what are the steps we need to consider this an implemented JEP?

SylvainCorlay:
> question for @SylvainCorlay - is this JEP (and the one for Xeus) now "implemented"? If not, what are the steps we need to consider this an implemented JEP?

I think this is mostly implemented. One small detail is that the link to gitter still points to the QuantStack gitter channel.

choldgraf:
Whenever you think this is done, we should mark them as completed in the JEP repository - want to make sure that whatever is supposed to happen when a project "becomes a jupyter proejct" actually happens 😅

PR #58, opened 2020-06-19
add workflow for link checking and spelling

### LINKED ISSUE #56: Check spelling, links of website HTML
Now that #53 is more formally publishing a website, additional measures would be a nice addition to achieve (and keep) high quality text, and reduce review time.

I don't know if these would be things that would need their own JEP, or are just nuts-and-bolts that fall out of (and would need documenting _anwyay_) in #29.

Notionally these would be run on every PR push.

## spelling
tool: `hunspell`

There are a bunch of other ones, but hunspell, can be installed via most system package manager or conda. windows is less rosy. 

I've found it best set up to test against generated HTML, where the semantics are more clear than markdown.... though would be happy to hear if anyone has something better.

While somewhat odius, maintaining a single `dictionary.txt` (just a sorted list of words, or a cleverer, but less portable stem-aware confection) across a corpus is good for keeping jargon and names consistent, e.g. `JupyterHub vs Jupyterhub`.

This also interestingly adds some more punch to the "new terms" part of the template.

## links
tool:  https://github.com/jupyterlab/pytest-check-links 

Used on other jupyter projects, though other options exist. We recently added some caching stuff to try to reduce the amount of rate limiting that was occurring for some public sites. The in-document anchor checking doesn't _quite_ work for sphinx output, but a fix has been merged, and could probably be released.

bollwyvl:
- fixes #56

This adds a new workflow which performs:
- spellchecking with `hunspell` against en-US and en-GB (the latter is more frequently updated)
  - this gets installed from `conda-forge`
  - **findings**
    - nothing too surprising: Jupyter, Xeus, etc. aren't in the dictionary... yet!
    - otherwise, awesome! good job team!
- link checking with `pytest-check-links`
  - this gets installed with `pip`
  - currently a bunch of declarativewidth
  - **findings** 
    - a bunch of the declarativewidgets links are broken. they just get skipped for now

I've tried some caching (still learning actions), and the total delta seems not so long. I didn't touch the existing workflows, but they should probably share requirements files. It has been suggested that we might get reusable step definitions (a la azure pipelines) in the future.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,99 @@
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - uses: actions/cache@v2
+      with:
+        path: ~/.cache/pip
+        key: pip-build-${{ hashFiles('.ci_support/requirements-build.txt') }}
+        restore-keys: |
+          pip-build-
+          pip-
+
+    - run: |
+        pip install -U -r .ci_support/requirements-build.txt
+
+    - run: |
+        jupyter-book build .
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: _build
+        path: _build
+
+  spelling:
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/download-artifact@v2
+        with:
+          name: _build
+          path: _build
+
+      - name: Cache conda
+        uses: actions/cache@v1
+        with:
+          path: ~/conda_pkgs_dir
+          key: conda-spelling-${{ hashFiles('.ci_support/environment-spelling.yml') }}
+          restore-keys: |
+            conda-spelling-
+            conda-
+
+      - uses: goanpeca/setup-miniconda@v1.6.0
+        with:
+          activate-environment: spelling
+          channel-priority: strict
+          environment-file: .ci_support/environment-spelling.yml
+          use-only-tar-bz2: true
+
+      - shell: bash -l {0}
+        run: |
+          bash .ci_support/check_spelling.sh
+
+  links:
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/download-artifact@v2
+        with:
+          name: _build
+          path: _build
+
+      - uses: actions/cache@v2
+        with:
+          path: .pytest-check-links-cache.sqlite
+          key: link-cache
+
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: pip-check-links-${{ hashFiles('.ci_support/requirements-check-links.txt') }}
+          restore-keys: |
+            pip-check-links-
+            pip-
+
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+
+      - run: |
+          pip install -U -r .ci_support/requirements-check-links.txt
+
+      - run: |
+          bash .ci_support/check_links.sh
--- a/.ci_support/check_links.sh
+++ b/.ci_support/check_links.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+CACHE_NAME=$(pwd)/.pytest-check-links-cache
+KNOWN_FAILS="\
+ibm \
+or declarativewidgets
+"
+
+cd _build/html
+
+
+pytest \
+  --check-links \
+  --check-links-cache \
+  --check-links-cache-name $CACHE_NAME \
+  --check-links-cache-expire-after 604800 \
+  --links-ext html \
+  -k "not ($KNOWN_FAILS)"
--- a/.ci_support/check_spelling.sh
+++ b/.ci_support/check_spelling.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "checking spelling..."
+
+cd _build/html
+
+hunspell \
+  -d en-GB,en_US \
+  -p ../../.ci_support/dictionary.txt \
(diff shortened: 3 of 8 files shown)


.github/workflows/check.yml
```diff
@@ -0,0 +1,121 @@
+name: Linting
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
```
bollwyvl, line 9: Adding this here, aside from doing _one last check_ on `master` (which would get a little more visibility... I guess badges would be better!), also ensures all the caches are created in that "context"... PRs can be use it as a [fallback](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) with `restore-keys`.

.github/workflows/check.yml
```diff
@@ -0,0 +1,121 @@
...
+        with:
+          python-version: 3.7
+
+      - name: Install dependencies
+        run: |
+          pip install -U -r .ci_support/requirements-check-links.txt
+
+      - name: Find broken links
+        run: |
+          bash .ci_support/check_links.sh
```
choldgraf, line 121: Another option is that we could use the link checker from Sphinx. Jupyter Book supports users calling it with `jupyter-book build . --builder linkcheck`. It's a little finnicky in my experience, but maybe that's true of all link checkers. Could simply the build process a bit?

.github/workflows/check.yml
```diff
@@ -0,0 +1,121 @@
...
+        with:
+          activate-environment: spelling
+          channel-priority: strict
+          environment-file: .ci_support/environment-spelling.yml
+          use-only-tar-bz2: true
+
+      - name: Find misspelled words
+        shell: bash -l {0}
+        run: |
+          bash .ci_support/check_spelling.sh
```
choldgraf, line 78: What is the likelihood that this is going to be triggered on a regular basis just because people are using non-standard words in JEPs etc? I'm a bit concerned that this is going to feel like a nagging bot that keeps telling people to slightly change wording that they think is correct. Do you know what I mean?

.github/workflows/check.yml
```diff
@@ -0,0 +1,121 @@
...
+        pip install -U -r .ci_support/requirements-build.txt
+
+    - name: Build the book
+      run: |
+        jupyter-book build .
+
+    - name: Upload book
+      uses: actions/upload-artifact@v2
+      with:
+        name: _build
```
choldgraf, line 42: wow I didn't realize it was so easy to persist artifacts in GHA haha

bollwyvl:
On Fri, Jun 19, 2020, 10:36 Chris Holdgraf <notifications@github.com> wrote:

> *@choldgraf* commented on this pull request.
>
> A couple quick comments in there - thanks for taking a stab at building
> some infra for QC!
> ------------------------------
>
> In .github/workflows/check.yml
> <https://github.com/jupyter/enhancement-proposals/pull/58#discussion_r442875875>
> :
>
> > +          restore-keys: |
> +            pip-check-links-
> +            pip-
> +
> +      - name: Set up Python 3.7
> +        uses: actions/setup-python@v1
> +        with:
> +          python-version: 3.7
> +
> +      - name: Install dependencies
> +        run: |
> +          pip install -U -r .ci_support/requirements-check-links.txt
> +
> +      - name: Find broken links
> +        run: |
> +          bash .ci_support/check_links.sh
>
> Another option is that we could use the link checker from Sphinx. Jupyter
> Book supports users calling it with jupyter-book build . --builder
> linkcheck. It's a little finnicky in my experience, but maybe that's true
> of all link checkers. Could simply the build process a bit?
>
- not sure if it caches: in ci, it's easy to get api limited. Gh in
particular.

- on another project with a heavily customized Sphinx build, I came to the
realization that i don't care how the link gets there, or what it is in
some intermediate form: all that matters is what ends up in the html, and
for all asset types, including those injected by themes, plugins, etc. that
might occur after the Sphinx link checker runs.

------------------------------
>
> In .github/workflows/check.yml
> <https://github.com/jupyter/enhancement-proposals/pull/58#discussion_r442876400>
> :
>
> > +          restore-keys: |
> +            conda-spelling-
> +            conda-
> +
> +      - name: Install dependencies
> +        uses: goanpeca/setup-miniconda@v1.6.0
> +        with:
> +          activate-environment: spelling
> +          channel-priority: strict
> +          environment-file: .ci_support/environment-spelling.yml
> +          use-only-tar-bz2: true
> +
> +      - name: Find misspelled words
> +        shell: bash -l {0}
> +        run: |
> +          bash .ci_support/check_spelling.sh
>
> What is the likelihood that this is going to be triggered on a regular
> basis just because people are using non-standard words in JEPs etc? I'm a
> bit concerned that this is going to feel like a nagging bot that keeps
> telling people to slightly change wording that they think is correct. Do
> you know what I mean?
>
High. But that's that point. Heavy use of unexplained jargon is not helpful
in a spec.  If a JEP introduces new terms, then they are new terms, and
there's a section for it called out in the template.  We could fiat all
such terms directly, but only if they are headings under the the section in
question. Xeus would be a good example of this.

We could grep out capital words/mixed case words, which would catch
names... But then a typo in a name elsewhere would not be caught. And that
makes it harder to catch Jupyterhub vs JupyterHub.

------------------------------
>
> In .github/workflows/check.yml
> <https://github.com/jupyter/enhancement-proposals/pull/58#discussion_r442876573>
> :
>
> > +        restore-keys: |
> +          pip-build-
> +          pip-
> +
> +    - name: Install dependencies
> +      run: |
> +        pip install -U -r .ci_support/requirements-build.txt
> +
> +    - name: Build the book
> +      run: |
> +        jupyter-book build .
> +
> +    - name: Upload book
> +      uses: actions/upload-artifact@v2
> +      with:
> +        name: _build
>
> wow I didn't realize it was so easy to persist artifacts in GHA haha
>
Yeah, it's decent... But not accessible between different workflow yaml
files. And they flake some time. And you can't "needs" them, just a whole
"job," which is kind of a bummer.

choldgraf:
yeah that makes sense - in that case I'm +1 on trying out this infrastructure and seeing how it feels in practice. Good point about the linkcheck cacheing and rate limits.

also just a note that when you reply to these via email, it's injecting a ton of extra stuff into your comments :-)

Issue #68, opened 2021-04-22
Pre-Proposal: Transmit Cell Metadata


jflam:
# Transmit Cell Metadata

Hello! 

We (@willingc and @MSeal) would like to propose a new JEP that describes transmitting cell metadata to kernels. Below is a summary and a link to the full proposal as well.

Thx! 
-John, Carol, Matt

## Summary

This proposal discusses the transmission of cell metadata with execute message requests and would modify the Jupyter Messaging Protocol. Individual kernels would interpret or ignore this metadata. This enables flexibility in different usage scenarios implemented in various front-end clients.

## Motivation

By transmitting cell metadata inline with the execute message request, Jupyter implementations will have a reliable channel to transmit additional metadata to the kernel in a standard way.

Notebook extensions can also use this channel to transmit additional information that was often transmitted using magic commands.

Some use cases which motivated this proposal are:

-   Route requests automatically to an appropriate kernel via libraries like [allthekernels]https://github.com/minrk/allthekernels) without need for additional metadata within the cell itself
-   Create or find a conda environment without needing to use magics, like [pick](https://github.com/nteract/pick)
-   Support polyglot (more than one language/kernel within a single notebook) scenarios, like [sos]https://vatlab.github.io/sos-docs/)
-   Provide hints to the kernel for localization purposes, like how the `ACCEPT_LANGUAGE HTTP` header works
-   Provide hints to the kernel about client capabilities, similar to how hints of a web browser client's capabilities work

A link to the full proposal can be found [here](https://github.com/jflam/enhancement-proposals/blob/cell-metadata/cell-metadata/transmit-cell-metadata.md)

blois:
Clients can already include metadata in execute requests- for example JupyterLab includes `cellId` and `recordTiming`. 

The scenarios outlined in the full proposal seem to require cooperation between the client and kernel- for example polyglot editors would want to support per-cell languages and the kernel needs to know how to handle the executions. What is preventing polyglot notebooks from passing the language/kernel alongside `cellId` and `recordTiming` today?

MSeal:
> Clients can already include metadata in execute requests- for example JupyterLab includes cellId and recordTiming.

[...]

PR #70, opened 2021-06-12
Proposal: Transmit Cell Metadata

jflam:
# Transmit Cell Metadata

Hello!

We (@willingc and @MSeal) would like to propose a new JEP that describes transmitting cell metadata to kernels. This PR contains the JEP proposal that we originally discussed in the [Pre-Proposal: Transmit Cell Metadata Issue](#68).

Based on the discussion in that issue, I've updated the JEP to reflect the desire to send the cell metadata as part of the message header vs. the original proposal which argued for sending it in the content. 

@minrk has graciously volunteered to shepherd this through the JEP process. Looking forward to continuing the discussion here!

Thx!
-John, Carol, Matt


choldgraf:
@jflam just FYI the failing test is because Jupyter Book updated its spec for the TOC, and we haven't pinned the versions in this repository. I'm updating the toc structure here: https://github.com/jupyter/enhancement-proposals/pull/71

zz-cell-metadata/transmit-cell-metadata.md
```diff
@@ -0,0 +1,347 @@
...
+
+# Reference-level Explanation
+
+Cell metadata will be transmitted to the kernel as part of the message
+metadata for the
+[execute](https://jupyter-client.readthedocs.io/en/stable/messaging.html#execute),
+[inspect_request](https://jupyter-client.readthedocs.io/en/stable/messaging.html#introspection),
+and
+[complete_request](https://jupyter-client.readthedocs.io/en/stable/messaging.html#completion)
+messages.
```
minrk, line 158: Add a note: In general, messages 'associated' with a cell should include the cell's metadata.

zz-cell-metadata/transmit-cell-metadata.md
```diff
@@ -0,0 +1,347 @@
...
+    //...
+  },
+  "parent_header": {},
+  "content": {},
+  "buffers": [],
+}
+```
+
+We propose the addition of a new `metadata` dict to the `header` dict schema
+in [nbformat](https://nbformat.readthedocs.io/en/latest/). This will be used
```
minrk, line 177: Some clarification:

1. message format is not specified in nbformat, it's in the [client docs](https://jupyter-client.readthedocs.io/en/stable/messaging.html)
2. there is already a `message.metadata` field, I don't think we don't need an extra `message.header.metadata` for the same thing

zz-cell-metadata/transmit-cell-metadata.md
```diff
@@ -0,0 +1,347 @@
...
+to transmit the cell metadata for the executed cell. 
+
+In cases where Jupyter extensions generate their own metadata, that keys for
+the metadata should be namespaced using an extension-specific prefix. The
```
minrk, line 181: It would probably be a good idea to recommend a sub-dictionary as well for this. There's no requirement (or even recommendation) that metadata be a one-level dict. If an extension has more than one or two settings, it probably makes sense to nest a dict.

zz-cell-metadata/transmit-cell-metadata.md
```diff
@@ -0,0 +1,347 @@
...
+```
+
+Below is the corresponding EXECUTE message:
+
+```js
+{
+  "header" : {
+    "msg_id": "...",
+    "msg_type": "...",
+    "metadata": {
```
minrk, line 217: I think this should be in the top-level `message.metadata.cell_metadata`, not the header.

minrk (commented):
Great! Based on #68, what I was thinking was to use the *existing* `message.metadata` field and add a `metadata.cell_metadata` (or `metadata.cell`) field there, rather than duplicating metadata into the header.

zz-cell-metadata/transmit-cell-metadata.md
jflam: I like this idea; do you think that existing implementations might make the incorrect assumption that there aren't nested dicts, i.e., this change would break existing implementations?

zz-cell-metadata/transmit-cell-metadata.md
jflam: Fixed

zz-cell-metadata/transmit-cell-metadata.md
jflam: Don't know what I was thinking when I added that. Must be tired :) Fixed!

zz-cell-metadata/transmit-cell-metadata.md
jflam: Fixed in newer commit. Thx!

zz-cell-metadata/transmit-cell-metadata.md
```diff
@@ -0,0 +1,348 @@
...
+`python3`. But where did the `"kernel": "python3"` metadata come from? What
+wrote it into the cell metadata in the first place? 
+
+Elaborating a bit more on the user experience here, you could imagine a client
+extension providing some additional UI elements such as a cell drop-down that
+lets the user pick from a list of installed kernels on the user's machine. The
+user picks one, and the kernelspec or its identifier is written to that cell's
+metadata.
+
+In this example, there is also a corresponding `allthekernels` kernel that is
```
damianavila, line 86: In this example you are supposing there is a corresponding kernel that "understand" the new metadata. What happens when this notebook finds a place where that kernel is not actually installed?

willingc: Excellent point @damianavila. We should be explicit about what happens if a kernel is not installed or able to support the messages. @jflam @codeseal I remember discussing this but I can't remember what we decided.

P.S. Thanks for reviewing @damianavila and @minrk.

zz-cell-metadata/transmit-cell-metadata.md
MSeal: At least in the python layer all of the library interactions within jupyter / nteract don't expect single layer dicts. I'd be surprised if the UIs did either since there's already multi-layer dicts present in nbformat fields.

zz-cell-metadata/transmit-cell-metadata.md
MSeal: So existing kernels would ignore the new field and operate as they do today, such that existing kernels that do not rely on the feature don't need to make any changes. If a new kernel starts using the feature it would ideally be a named kernel (e.g. `allthekernels` kernel) that would then be saved into the metadata of the notebook as the kernel being used.

One could name the kernel a generic name that overlaps with a non-supporting kernel, or a fallback mechanism might use the kernel language info to incorrectly use a non-supporting kernel. In those cases the notebook session would most likely spew a bunch of language errors to the user about that Powershell code not being valid .NET syntax. When we discussed this I recall us pointing out that this problem already exists with weak kernel id matching and it would be out of scope of this change to be solved. Specifically the problem of picking an incompatible kernel is already present and not made worse by adding new kernel features.

I believe that out-of-scope problem was discussed around [dependency tracking](https://github.com/nteract/nes/blob/master/dependency-management/README.md) as well in the past, so it's been a topic of discussion that warrants digging into again down the road imo.

zz-cell-metadata/transmit-cell-metadata.md
westurner: There should be nothing that prevents transmission (and a space-limited journal; storage) of nested dicts i.e. with schema: JSONschema-validateable [nested] JSON and/or W3C SHACL-validateable JSON-LD with URIs for migrateable Linked Data.

minrk: I don't think there's any risk associated with it. Dicts are generally nested throughout Jupyter and the only spec for metadata is that it's a JSONable dict, which can have aribtrary depth, just like is already true for header, content, etc. which can also have nested fields.

zz-cell-metadata/transmit-cell-metadata.md
damianavila: >Specifically the problem of picking an incompatible kernel is already present and not made worse by adding new kernel features.

Maybe not worse _per se_, maybe even better because we will see the problem more frequently and that will "force" us to fix it down the road, as you mentioned.

Thanks for the feedback @willingc and @MSeal.

willingc:
Hi folks,

This PEP has been open for 3 weeks. We'll be collecting feedback for another week. On Monday July 12, we will request a vote.

Thanks!

Carol

meeseeksmachine:

https://discourse.jupyter.org/t/jep-70-transmit-cell-metadata/9877/1

jasongrout (commented):
At first glance, I thought this proposal was adding a new field to the message protocol, but a closer reading indicates this understanding was wrong - it is using the existing metadata field all messages have. Is that correct?

From my understanding, thinking as a person that will need to implement this JEP and exactly what would need to be done, what this proposal is proposing is the following. All Jupyter clients that are executing code from the notebook format (i.e., there is a notion of "cell" and "cell metadata" in the client), when they are working with a "cell", must send all cell metadata from the notebook format to the kernel in any `execute_request`, `inspect_request`, or `complete_request` messages. This cell metadata is merged into the top-level metadata dictionary of the above message. Is this a correct understanding? Is there anything else that would need to be implemented if this JEP is passed?

In other words, since the metadata field already exists in these messages, a frontend can already transmit this information if it wishes, so presumably a frontend and kernel can coordinate on this metadata already. Is this JEP really about *mandating* that the metadata is transmitted from the frontend?

jasongrout:
Another few questions trying to understand the scope of this proposal:

1. Is there anything that frontends without the notion of a "cell" or "cell metadata" do? For example, the ipython console or JupyterLab Code Console?
2. What about comm messages? Often ipywidgets sends comm messages that are sort of triggered from a cell - do they need to send cell metadata? What about debug messages?
3. Right now, as the JEP says:
    > The attributes within the metadata today are: A) small in size and B) not harmful to send across the wire so keeping the solution simpler was the preferred pattern in the proposal.
   
   Of course, we don't know what custom extensions are putting in cell metadata today, and it seems limiting to make these assumptions about all cell metadata for the future as well. I would strongly suggest a specific place for cell metadata that is meant to be sent to the kernel, rather than always sending everything. For example, we could have a new top-level dictionary for cell metadata to be sent to the kernel. Or perhaps a white-list of fields to send can be in the notebook metadata, so extensions can add metadata and tell the frontend to send it?
4. One of the motivation examples includes providing hints about client capabilities. It seems awkward if each cell has repeated metadata about the frontend capabilities. For that usecase, is it better to think about this metadata as "any data the frontend wants to send the kernel" rather than just "cell metadata"?

jasongrout:
>     4\. One of the motivation examples includes providing hints about client capabilities. It seems awkward if each cell has repeated metadata about the frontend capabilities. For that usecase, is it better to think about this metadata as "any data the frontend wants to send the kernel" rather than just "cell metadata"?

I agree with Min's thoughts [here](https://github.com/jupyter/enhancement-proposals/issues/68#issuecomment-828298718), that we should separate cell metadata from other types of metadata that is transmitted.

jasongrout:
>  2\. What about comm messages? Often ipywidgets sends comm messages that are sort of triggered from a cell - do they need to send cell metadata? What about debug messages?

Min's thoughts [here](https://github.com/jupyter/enhancement-proposals/issues/68#issuecomment-825914333) indicate we should think about this as a generic way of associating cell metadata with any message, so the answers to the above would generally be "yes", i.e., if the message is closely associated with a cell, the relevant metadata should be transmitted.

jasongrout:
> In other words, since the metadata field already exists in these messages, a frontend can already transmit this information if it wishes, so presumably a frontend and kernel can coordinate on this metadata already. Is this JEP really about _mandating_ that the metadata is transmitted from the frontend?

I think it would be very clarifying to see exactly what changes are going to be made to the [message spec document](https://jupyter-client.readthedocs.io/en/stable/messaging.html) as part of this JEP, e.g., are there new prescribed fields in the message metadata dict, what their purpose is, etc.

minrk:
@jasongrout thanks for the review!

Yes, I [think](https://github.com/jupyter/enhancement-proposals/issues/68#issuecomment-828298718) we should specify that cell metadata SHOULD be transmitted as-is in `message.metadata.cell` for any message reasonably associated with a cell (rather than directly at top-level `message.metadata` so there's room for e.g. `client` metadata or others without collisions).

jflam:
Thanks for the thoughtful comments, @jasongrout!

>> One of the motivation examples includes providing hints about client capabilities. It seems awkward if each cell has repeated metadata about the frontend capabilities. For that usecase, is it better to think about this metadata as "any data the frontend wants to send the kernel" rather than just "cell metadata"?

> I agree with Min's thoughts here, that we should separate cell metadata from other types of metadata that is transmitted.

I think you've done a great job at identifying a key tension in this JEP: *cell* metadata vs. *client* metadata. Looking at Min's comments from above, what do folks think about having two different fields in the message metadata, i.e., `message.metadata.cell` and `message.metadata.client`? This would make the intent clearer at the cost of additional complexity. There are other optimizations possible such as transmitting `client` metadata only if an actual change was initiated by the client, for example.

> Is there anything that frontends without the notion of a "cell" or "cell metadata" do? For example, the ipython console or JupyterLab Code Console?

One of the primary scenarios for transmitting cell metadata is identifying the programming language in an `execute_request` message. In this case, I wonder if there isn't a list of reserved/standard messages that need to be there so that clients like the ones you describe above will know what message to transmit to the kernel, rather than that being an implicit contract between a front end and the kernel?

> Right now, as the JEP says:

>>The attributes within the metadata today are: A) small in size and B) not harmful to send across the wire so keeping the solution simpler was the preferred pattern in the proposal.

> Of course, we don't know what custom extensions are putting in cell metadata today, and it seems limiting to make these assumptions about all cell metadata for the future as well. I would strongly suggest a specific place for cell metadata that is meant to be sent to the kernel, rather than always sending everything. For example, we could have a new top-level dictionary for cell metadata to be sent to the kernel. Or perhaps a white-list of fields to send can be in the notebook metadata, so extensions can add metadata and tell the frontend to send it?

Creating an allow-list in this manner, allowing the notebook to specify it is a different take than what we originally rejected with the kernel being the source of truth. It does have the downside of additional complexity in implementation. Thinking out loud: if there is an explosion in the size of cell metadata down the road, we could mandate that requirement in a future JEP as an optimization that clients would be eager to implement. This would give us a simpler implementation for this JEP and give us time to gather evidence for the more complex implementation. What do you think?

jasongrout:
> One of the primary scenarios for transmitting cell metadata is identifying the programming language in an `execute_request` message.

Here's a thought: instead of transmitting all cell metadata, or even a language name, how about we take a cue from our existing mechanism for identifying rich output formats, and instead transmit a mimetype field that identifies the format of code being transmitted? The mimetype of a typical python execute request could be `text/x-python`, for example. This naturally opens the door for other kinds of input to even single-language kernels, for example a special mimetype that represents a JSON datastructure for setting values in a language runtime, or other mimetypes that represents a chunk of binary data you are sending to the kernel (like an image or parquet file). Perhaps a sql kernel usually has text sql input, but one of the inputs could be a JSON datastructure containing info about what database to connect to (with a mimetype specific to that format) - that db info is sent in an execute_request with a mimetype of `application/vnd.sqlkernel.dbinfo+json`.

I realize it's basically the same thing as a language field - a string that identifies to the kernel how to handle the execution/completion request. I usually think of a kernel as having a single "language" (this is implicit in the kernel spec, where the kernel has a "language" identified), but it seems okay to say that a given kernel can receive and act on multiple types of inputs.

jflam:
Yes- using a MIME type to identify a language would make perfect sense! In the prior art section of the JEP, I do mention a few other "meta-kernels" that exist as well that would benefit from this mechanism.

I do like the examples that you give, and it seems like some of those would be metadata that is not likely to be persisted in the .ipynb itself (e.g., sending chunk of binary data) and some that are likely to be persisted, (e.g., setting values in the language runtime - of course depends on what those values represent ...). I wonder if there's a more general concept "runtime metadata?" where some of the metadata comes from cell metadata and others are synthesized by the front end/extensions at runtime?

meeseeksmachine:

https://discourse.jupyter.org/t/propagating-user-environment-information-to-ipython-kernels-and-more-generally/10931/1

meeseeksmachine:

https://discourse.jupyter.org/t/notebook-cell-type-generaliastion/10703/1

PR #77, opened 2021-10-09
Add write permission for the bot

### LINKED ISSUE #76: The JEP website is outdated, recent CI jobs fail
Looking at https://github.com/jupyter/enhancement-proposals/runs/3689450973 the relevant part is:

```
remote: Permission to jupyter/enhancement-proposals.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/jupyter/enhancement-proposals.git/': The requested URL returned error: 403
```

krassowski:
This may fix #76, but ideally the write permission would be restricted to `gh-pages`. Not sure if this is possible.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,6 +6,10 @@ on:
     branches:
     - master
 
+# Allow the bot to push to the repository
+permissions:
+  contents: write
+
 # This job installs dependencies, build the book, and pushes it to `gh-pages`
 jobs:
   deploy-book:


choldgraf:
Let's give this a shot and see if it fixes the problem 🙂 many thanks for the PR!

I think we could prevent pushing to main/master by turning it into a protected branch

krassowski:
Thanks for the merge, it seems to be working :tada:.

> I think we could prevent pushing to main/master by turning it into a protected branch

I think that's right, that should definitely help.

Issue #78, opened 2021-11-19
JEP: Notebook v7


fperez:
The transition from the version 6 of the notebook to a future version 7, based on the modern JupyterLab codebase, has been already extensively discussed in our community. Given there have been multiple open issues (listed below), meetings and discussions about this already, I'm shortcutting a bit here by moving directly to creating a JEP PR for this topic. But in case there's any concerns folks would like to address here outside of discussing the JEP proper, I'm opening this issue for the time being.

To keep things focused, any material discussion of the JEP body itself should go into the PR, not here.

Previous issues of relevance:

* [[Discuss] The future of the classic notebook interface](https://github.com/jupyter/notebook/issues/6210).
* [Proto-JEP](https://github.com/jupyter/notebook/issues/6220).
* [Early draft of the JEP in my fork, that was edited by all the JEP co-authors](https://github.com/fperez/enhancement-proposals/pull/1).

jasongrout:
The JEP has been accepted and the JEP PR merged (#79).

PR #79, opened 2021-11-19
Jupyter Notebook version 7

fperez:
### Summary

This PR proposes a path forward for the _user experience_ of the `jupyter notebook` application to continue to be available to users, with the upcoming version 7 based on the JupyterLab TypeScript codebase and APIs. Version 7 of the notebook will use what is today [RetroLab](https://github.com/jupyterlab/retrolab), with suitable repo restructuring as detailed in the JEP.

In #78 I created the issue, but as indicated there, I'm moving directly to making a PR, given the extensive amount of community discussion this topic has already received. This PR should be where specifics of the proposal itself and Notebook v7 plan are discussed. Any meta/process issues can be dealt with in #78, if need be.

### References

For reference, these are the main  issues where this topic had been previously discussed:

* [[Discuss] The future of the classic notebook interface](https://github.com/jupyter/notebook/issues/6210).
* [Proto-JEP](https://github.com/jupyter/notebook/issues/6220).
* [Early draft of the JEP in my fork, that was edited by all the JEP co-authors](https://github.com/fperez/enhancement-proposals/pull/1).
* [Notes for the JupyterLab team meetings](https://github.com/jupyterlab/team-compass/issues/128#issuecomment-971829367).
* [RISE: new approach for JupyterLab 3](https://github.com/damianavila/RISE/pull/605).

In addition, see [the Enhancement Proposal guidelines](https://jupyter.org/enhancement-proposals/jupyter-enhancement-proposal-guidelines/jupyter-enhancement-proposal-guidelines.html#goals-of-the-jep-process) for information about the goals of the process itself.

### Discussion points

- ✔️Reassurances on stability for extension authors ([link](https://github.com/jupyter/enhancement-proposals/pull/79#pullrequestreview-813223863))
- ✔️ Should this JEP include the scope of "Jupyter Desktop" as well? ([link](https://github.com/jupyter/enhancement-proposals/pull/79#discussion_r753711278))
- ✔️ What do do with RetroLab? ([link](https://github.com/jupyter/enhancement-proposals/pull/79#discussion_r754024876))

### Voting from the @jupyter/steeringcouncil 

- @afshin
  - [x] Yes
  - [ ] No
- @blink1073
  - [x] Yes
  - [ ] No
- @Carreau
  - [x] Yes
  - [ ] No
- @damianavila
  - [x] Yes
  - [ ] No
- @ellisonbg
  - [x] Yes
  - [ ] No
- @fperez
  - [x] Yes
  - [ ] No
- @ivanov
  - [ ] Yes
  - [x] No
- @jasongrout
  - [x] Yes
  - [ ] No
- @jhamrick
  - [ ] Yes
  - [ ] No
- @minrk
  - [x] Yes
  - [ ] No
- @mpacer
  - [ ] Yes
  - [ ] No
- @parente
  - [ ] Yes
  - [ ] No
  - [x] Abstain
- @rgbkrk
  - [x] Yes
  - [ ] No
- @Ruv7
  - [ ] Yes
  - [ ] No
  - [x] Abstain
- @SylvainCorlay
  - [x] Yes
  - [ ] No
- @takluyver
  - [ ] Yes
  - [ ] No
- @willingc
  - [x] Yes
  - [ ] No


fperez:
Pinging co-authors Sylvain Corlay (@SylvainCorlay), Afshin Darian (@afshin), Sharan Foga (@sharanf), Kevin Goldsmith (@KevinGoldsmith), Brian Granger (@ellisonbg) , Jason Grout (@jasongrout), Zach Sailer (@Zsailer), Jeremy Tuloup (@jtpio) that the proposal is ready!

Folks - I reordered all names alphabetically, but kept myself first as I didn't want to pop one of you "in front" (that would be @SylvainCorlay given the current names we have, unless a new co-author comes in before Corlay) without your authorization. But I'm making zero claims of authorship here, and I'm perfectly happy moving my name into the P slot as long as everyone, and particularly --at least for now-- Sylvain, are OK with being in front :)

fperez:
Note that I edited the PR title to reflect we should allow for some public input in this full version before moving on to a vote. While we've had a lot of discussion that I think warrants putting it here, the finished document only came together this morning, and the community should have a chance to mull it over and provide input so we have a solid plan moving forward with broad support.

choldgraf:
Hey all - just to make it explicit, and to match the title, I've converted this to a "DRAFT" in github's UI as well

afshin:
> I reordered all names alphabetically, but kept myself first ...

I am happy with the way you have it right now. Thanks, Fernando!

meeseeksmachine:

https://discourse.jupyter.org/t/jep-draft-open-for-the-notebook-v7-transition/11769/1

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,134 @@
...
+- Real-time collaboration
+- Theming and dark mode
+- Internationalization
+- Improved Accessibility
+- Support for many JupyterLab extensions, including Jupyter LSP (Language Server Protocol) for enhanced code completions
+
+## Technical details
+
+- [RetroLab](https://github.com/jupyterlab/retrolab) running on [Jupyter Server](https://github.com/jupyter-server/jupyter_server) will be the basis for Notebook v7.
+- The current [RetroLab](https://github.com/jupyterlab/retrolab) commit history will be grafted into the [Jupyter Notebook repository](https://github.com/jupyter/notebook) in a pull request.
```
manics, line 62: Is it worth mentioning what will happen to the retrolab pypi package and repo, e.g. will they be deprecated?

fperez: I'll defer to @jtpio - it seems to me that when the time comes for the transition, the simplest thing to do would be to tag a final release under the retro name and then end-point the project at that stage. From then on, `install notebook` will be the way to get that tool. We probably will want to update the project pages on pypi/conda to make that clear though.

But before editing the JEP, let's hear from Jeremy and others - the above is just my first thinking...

fperez:
I would like to suggest @choldgraf as shepherd for this JEP - Chris has worked a lot on the JEP process itself, has enough knowledge about the notebook and our community, but is reasonably removed from the specific decisions to provide the right balance of knowledge and impartiality, I think. I consulted and he's OK with the idea.

[ I'd originally posted this in the top summary, but best to have it here as a standalone comment for the record/discussion, as we may update the description in-place ]

choldgraf:
I am happy and willing to steward this JEP, as long as others are 👍 as well!

NN-notebook-v7/notebook-v7.md
SylvainCorlay: Thanks for bringing this up @manics. Since retrolab is part of the JupyterLab org, this JEP should probably say something about it, even if the solution is simple.

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,134 @@
...
+- As a user, I want the most common notebook extensions to work in both the notebook and lab interfaces, so I can move back and forth without missing functionality (see below).
+-   As a user, I want this user experience to work seamlessly with the [Jupyter Desktop](https://blog.jupyter.org/jupyterlab-desktop-app-now-available-b8b661b17e9a)
+    - As a user, I can double-click on an `.ipynb` file on my local computer (Mac, Win, Linux) to open the notebook as a single document which I can edit, run, etc.
+    - As a user, I want to be able to go from the document-centric experience to the full JupyterLab Desktop application, so I can get a more IDE-centric experience, including file management, terminals, and more.
+    - As a user, I can open notebooks via the command-line or open JupyterLab in a particular directory.
+    - As a user, I want Jupyter Desktop to let me open my Jupyter editor (with any extensions I may have installed) as a single application and the Jupyter Console.
+    - As a user, I can use other Jupyter-based tools in the native desktop experience, such as Consoles attached to Notebooks, in their own Native window.
```
krassowski, line 98: Do users want to have both `JupyterLab Desktop` and `Jupyter Notebook v7 Desktop`, or a single `Jupyter Desktop` app which will ship both `JupyterLab Desktop` and `Notebook Desktop`?

This is a question about organization of the future work on `JupyterLab Desktop`. While it is already possible to launch classic notebook in a separate window from JupyterLab Desktop (using `Help` menu), making the Notebook v7 a first class citizen sibling application could make it more attractive for some users.

I believe that the shared codebase will make it easier to make JupyterLab Desktop ship both experiences. Potentially we could brand the end product "Jupyter Desktop" (so it would include Desktop/launcher icons for both, say like Office includes both Word and Excel). I can see many technical benefits for bundling the two together (as opposed to duplicating the maintenance work in two separate repositories, or even shipping two installers from the same repository). The question is what do users want?

afshin: The `notebook` and `jupyterlab` packages will include both user interfaces and give the users the ability to switch between them, so Jupyter Desktop should have this same behavior, I would say.

ellisonbg: One question I have about the desktop notebook experience is this: will each notebook open as a separate OS window, or a new browser tab (I don't know if Electron apps have the equivalent of browser tabs). My sense is that separate OS windows makes the most sense.

kevingoldsmith: This change seems to be straying from the original JEP in scope, adding in Desktop Notebook and JupyterLab features. Maybe exclude those from the Notebook v7/Retrolab discussion? The first bullet is specifically missing the point of a lot of the "notebook classic" discussion from users.

NN-notebook-v7/notebook-v7.md
jtpio: > it seems to me that when the time comes for the transition, the simplest thing to do would be to tag a final release under the retro name and then end-point the project at that stage

Yes this sound like the simplest approach.

In this proposal we expect a release of Notebook 7 in 2022. But pre-releases can be done earlier. We can start deprecating the use of the `retrolab` package when pre-releases of Notebook 7 are being published, and encourage folks to install it with `pip install --pre notebook`.

The `retrolab` repo can be archived when Notebook 7 final is released.

NN-notebook-v7/notebook-v7.md
choldgraf: I had the same question as @kevingoldsmith - as steward it's not my place to say what _should_ be the scope of this proposal. But I wonder what @fperez and the JEP authors think about removing the specifics about "Jupyter Desktop" from this JEP specifically, in order to focus on the transition from v6 to v7. Is the "open on the desktop" story a crucial part of this discussion?

choldgraf:
I've cleaned up the top comment a little bit to create more natural sections of information. I'm also going to try an experiment and provide links to major ongoing discussions, to make it easier for people to follow where the major discussion points are.

NN-notebook-v7/notebook-v7.md
choldgraf: Can somebody propose a language update to the PR, or push a commit that updates the language, and we can resolve this?

NN-notebook-v7/notebook-v7.md
afshin: I think it's reasonable to remove the entire Jupyter Desktop section.

Jupyter Desktop is hugely important and valuable but I don't think it needs a JEP and I agree that it doesn't need to be part of *this* JEP.

NN-notebook-v7/notebook-v7.md
afshin:
```suggestion
- The current [RetroLab](https://github.com/jupyterlab/retrolab) commit history will be grafted into the [Jupyter Notebook repository](https://github.com/jupyter/notebook) in a pull request. When pre-releases of Notebook v7 are published, the `retrolab` package can be deprecated in preparation for archival after Notebook v7 final is released.
```

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,134 @@
...
+    - As a user, I want Jupyter Desktop to be able to authenticate to and connect to any local or remote Jupyter Server, so I can have the same native Desktop experience against remote resources.
```
afshin, line 99:
```suggestion
```

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,134 @@
...
+- This JEP was originally posted by @fperez [as a comment in the original discussion opened by @zsailer](https://github.com/jupyter/notebook/issues/6210#issuecomment-957169113).
+- The blockers for transitioning Berkeley's Data 8 course to RetroLab are tracked [here](https://github.com/berkeley-dsep-infra/datahub/issues/2422).
+- The [community-contributed Notebook extensions have their own repository](https://github.com/ipython-contrib/jupyter_contrib_nbextensions).
+
+## Questions
+
+**Q:** What implications does this transition have for the notebook `.ipynb` format?
+**A:** _None_. The file format remains unchanged. This JEP is about the Jupyter Notebook application, not the Jupyter notebook `.ipynb` file format.
+
+**Q:** I wrote a Notebook extension that I'd like to update. Where can I find resources and/or support for that?
```
amit1rrr, line 129: Migrating extensions is going to be the most cumbersome & long drawn part of this transition (educate community, provide resources, motivate etc.). To that end, is the current JupyterLab API more or less stable? Is the new distribution mechanism of _prebuilt extensions_ going to the be default long term?

The worst outcome would be for us to spend huge effort in helping people migrate their extension from Notebook V6 to V7 and the next major versions of JupyterLab changes extension authoring process (requiring re-migration). I'm assuming future versions of JupyterLab (V5, V6, V7) will be kept in sync with Notebook (V7, V8, V9).

Overall, this is a great JEP. We're modernising infra while keeping classic UX largely the same. Fantastic. My only point here is, if there's any extension authoring related improvements in JupyterLab, those should happen before this.

NN-notebook-v7/notebook-v7.md
SylvainCorlay: I merged @afshin's suggestion.

NN-notebook-v7/notebook-v7.md
SylvainCorlay: I merged @afshin's suggestion about retrolab's deprecation after notebook 7 is out.

NN-notebook-v7/notebook-v7.md
choldgraf: It feels like there are two parts to comment that are worth disentangling:

### Will major versions of JupyterLab and Jupyter Notebook be released simultaneously?

This makes sense to me because if a lab version bump introduces a breaking change to extensions, then we'd need a corresponding version bump in the notebook as well. Otherwise you'd need to ask for each Jupyter Lab major release if that also introduced a breaking change that would force the Jupyter Notebook to bump the version as well. It seems easier to just assume that they move together. Is that right?

### Should this JEP make some commitment to API stability of JupyterLab extensions?

I can see two sides to this one. On one hand, as @amit1rrr notes, if we ask people to do a bunch of migration work, they may want the knowledge that this will not be an ongoing chore that they have to keep putting time into as the API changes. On the other hand, I suspect it would be difficult to make any strong commitments about API stability at this moment. But maybe the JEP can include some basic language to assuage people's fears about having to constantly update their extensions?

I'll await feedback or commits from the JEP authors (or discussion here from others)

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,131 @@
...
+* **Option A**: Making Notebook V7 work and look exactly like Notebook v6
+* **Option B**: Making Notebook V7 work and look exactly like JupyterLab V4
+
+There will always be tradeoffs between these two options. To help us make these tradeoffs, we propose to look at the following questions:
+
+* Is Option A or Option B be more accessible?
+* If we were to design this from scratch today, would our design work and look more like Option A or Option B?
+* Is Option A or Option B more maintainable?
+* Is Option A or Option B be more secure?
+* Will Option A confuse or cause friction Jupyter users who work in both Notebook V7 and JupyterLab v4?
```
bollwyvl, line 111: _cause friction **to** Jupyter users_

NN-notebook-v7/notebook-v7.md
bollwyvl: > is the current JupyterLab API more or less stable?

A major change, likely to land in Lab v4, is the deprecation of the home-grown evented contents model for the `yjs`-backed one which enables collaborative editing.

There is also an on-going effort toward migrating some common TypeScript-native APIs and into JSON Schema-constrained ones: this has already occurred, for example, in the menu system. In these cases, the dynamic APIs will continue to be available, but being data-driven provides opportunity for different layers of the system (e.g. server, kernel, hub) to participate in UI customization.

> Is the new distribution mechanism of prebuilt extensions going to be the default long term?

The JupyterLab experiment has shown us that: 

- leveraging and integrating existing third-party tooling provided by `nodejs` is probably not optional
  - "demanding" `requirejs`-compatible distributions of upstream packages on `bower` was tiresome, and most extension managers ended up having to use `npm` and `webpack` anyway
- `nodejs` is not something we can depend on working day-to-day for the "average" Jupyter end user (whoever that is)
  - more concretely, `webpack` and `npm` create a lot of problems outside of non-modern Linux systems... and even sometimes there

To this end, a discussion is currently under way regarding future work in JupyterLab and its user-facing `nodejs`-based server, CLI, and UI components (e.g. the extension manager)
https://github.com/jupyterlab/jupyterlab/issues/11336

In all likelihood these capabilities (and APIs) would be maintained, but deprecated, in Lab 4, and then become a separately installed, developer-focused package at some time in the future.

This would leave the pre-built extension distribution technique, enabled by WebPack v5, as the only supported extension mechanism in "vanilla" Lab v5. At present, we have yet to find tooling that enables uncoordinated parties to co-create JS extensions to share the same DOM and can handle deduplication and/or simultaneous dependency versions. 

If an IETF/W3C standard for this kind of metadata around ES modules came about, broadly implemented by the major browsers, we would jump at the opportunity to embrace it. But until then, the momentum behind WebPack v5 (and no doubt an eventual v6) means that choice is unlikely to change.

And indeed: another outcome of the above might be transitioning _much_ more of the core Lab functionality into the pre-built extension mechanism. The practical impact here would be that a system with `notebook` (and by dependency, a compatible `jupyterlab`) would have exactly _one_ copy of, for example, `codemirror`, `ajv`, and `react`, in its `$PREFIX/share/jupyter/labextensions` folder. This is in contrast to today, where `retrolab` and `jupyterlab` ship two, separate copies of all of the above.

choldgraf: @bollwyvl is it correct to say that **notebook extension authors who wish to migrate should use the pre-built extension setup in JupyterLab, and that this will remain relatively stable for the foreseeable future**?

NN-notebook-v7/notebook-v7.md
SylvainCorlay, line 111:
```suggestion
* Will Option A confuse or cause friction to Jupyter users who work in both Notebook V7 and JupyterLab v4?
```

NN-notebook-v7/notebook-v7.md
jtpio: I think this statement is correct. We generally encourage extension authors to develop their extensions as prebuilt extensions since this greatly improves the user experience.

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Major new features in Notebook v7
+
+In addition to having parity with the Notebook v6 user experience and widely-used extensions, Notebook v7 will inherit major new features that have been developed in the JupyterLab project, thus providing automatically a number of improvements to Notebook users as they transition to v7:
+
+- Debugger
+- Real-time collaboration
+- Theming and dark mode
+- Internationalization
+- Improved Accessibility
```
tonyfast, line 56:
```suggestion
- Improved Web Content Accessibility Accessibility compliance
```

it might help to be a bit more specific in the gains on accessibility. our initial goals will be to satisfy WCAG compliance before diving into more bespoke aspects of accessibility.

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Major new features in Notebook v7
+
+In addition to having parity with the Notebook v6 user experience and widely-used extensions, Notebook v7 will inherit major new features that have been developed in the JupyterLab project, thus providing automatically a number of improvements to Notebook users as they transition to v7:
+
+- Debugger
+- Real-time collaboration
+- Theming and dark mode
+- Internationalization
+- Improved Web Content Accessibility Accessibility compliance
```
afshin, line 56:
```suggestion
- Improved Web Content Accessibility Guidelines (WCAG) compliance
```

NN-notebook-v7/notebook-v7.md
choldgraf: Sounds good - before resolving this thread, I'll await guidance from the JEP authors about whether they'd like to:

1. Incorporate language in the JEP that addresses the conversation here
2. Treat this as an implementation detail and out-of-scope for the JEP

afshin: I think it's fair to leave this discussion out of the JEP on the good faith assumption that the API changes to come between major versions will be on the same order of magnitude as any normal API change between major versions. 

The pre-built extension system was a hard-won piece of engineering and it is here to stay. User-land APIs will change over time as the project evolves, but those changes will be well-documented.

Extension authors are one of the most important constituencies of the Jupyter community and we do not want to leave anybody behind from one major version to the next.

cf.
* [JupyterLab 1.x => 2.x migration guide](https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#jupyterlab-1-x-to-2-x)
* [JupyterLab 2.x => 3.x migration guide](https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#jupyterlab-2-x-to-3-x)
* [JupyterLab 3.0 => 3.1 migration guide](https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#jupyterlab-2-x-to-3-x)

choldgraf:
I thought it would be useful to make this JEP follow the standard template in the enhancement proposals guidelines. Here's a PR that rearranges the content here to match what is in the template: https://github.com/fperez/enhancement-proposals/pull/3

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,131 @@
...
+
+**Q:** What implications does this transition have for the notebook `.ipynb` format?
+**A:** _None_. The file format remains unchanged. This JEP is about the Jupyter Notebook application, not the Jupyter notebook `.ipynb` file format.
+
+**Q:** I wrote a Notebook extension that I'd like to update. Where can I find resources and/or support for that?
+**A:** We will strive to assist the community throughout the transition to the new extension system. There will be a central location with information on how to port extensions for others to do so in the future. We will also host office hours to assist extension developers.
+
+**Q:** What will I need to do to get my JupyterLab extension working with Notebook v7?
+**A:** Both JupyterLab and Notebook v7 will use the same extension system and it is our intention to enable extension authors to create extensions that work in both JupyterLab and Notebook v7. We are still working out the technical details of that and will update our documentation with links as more information is available.
+**A:** ...
```
SylvainCorlay, line 131:
```suggestion
```

choldgraf:
### Proposal

Given that we've already had a lot of conversation in the threads linked in the top comment, and a few iterations here, I suggest that we try to move this forward sooner than later.

Unless there are objections, I suggest we do the following:

- **Leave the RFC period open for 48 hours**.
- After 48 hours, if there are no objections to moving forward, the JEP authors resolve remaining comments and make edits to the text as needed.
  - If somebody objects to moving forward, please suggest an alternative timeline or criteria for moving forward.
- At this point, the JEP authors should feel free to trigger a vote when they wish.

afshin:
I am 👍 on your plan, Chris.

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,127 @@
...
+
+- [nbgrader](https://github.com/jupyter/nbgrader)
+- [RISE](https://github.com/damianavila/RISE)
+- [Jupytext](https://github.com/mwouts/jupytext/issues/271)
+- IPython Parallel (filebrowser tab)
+
+### Extensions whose functionality will be built into Notebook v7
+
+The following extensions will have largely equivalent experiences shipping in core Notebook v7 and JupyterLab. We will need to track whether there are any remaining differences in their functionality that are either regressions, or potentially disrupting or confusing to existing users.
+
```
ellisonbg, line 41: I think we should add the other JupyterLab extensions that we envision working in retrolab and that require significant changes such as the debugger, property panel, status bar, etc.

SylvainCorlay: Since these extensions (such as the debugger) don't exist in the classic notebook, maybe we don't need to be more specific about this as part of this JEP.

afshin: I think those pieces of functionality will be important as part of the blog posts, community announcements, marketing pitch, etc. to market Jupyter Notebook v7 to end users, but I agree with Sylvain that it might not be necessary for the JEP itself. Do you think it's critical to the proposal itself, Brian?

choldgraf: My 2 cents: I think that @SylvainCorlay and @afshin make reasonable points here - this JEP feels to me more like "bringing the notebook into the JupyterLab ecosystem", rather than "bringing JupyterLab's experience into the notebook". I feel like the target audience is the community of people who use the classic notebook interface, and want to make sure their *current* notebook workflows are not disrupted too much. For that reason, I don't think that defining the pre-existing Lab extensions that will work with Notebook v7 is on the critical path to getting this JEP merged. Though if somebody wishes we could always just add some informal list items to the bullet point here:

https://github.com/jupyter/enhancement-proposals/pull/79/files/30129955b28a6f7362f3dcf616c4034bc804a53d#diff-bf4dfc36e69ce75feec397491bcdd9572c0de9c663a4d5e77303a1c812c5348aR54

ellisonbg: I am fine with the JEP moving forward without any changes to the document, but we mention in other section the new features that notebook 7 will get by being based on JupyterLab and things like the debugger are mentioned. But these details are tactical and implementation related that don't need to be in the JEP.

SylvainCorlay: Great, let's resolve this note and move this JEP to the next stage!

NN-notebook-v7/notebook-v7.md
choldgraf: OK I'll resolve this one, as I think that we've decided that committing to a specific JupyterLab extension stability is not in-scope for the JEP, though it also sounds like the general guidelines will be

> use the federated extension approach, and subsequent changes to JupyterLab's extension API will likely be minor (compared to the original extension migration) and will come with documentation and a migration guide.

I think it's more important that language like this makes it into the extension migration docs, but probably not needed for this JEP! Thanks all for the discussion

choldgraf:
Hey all - I believe that all major outstanding comment threads are resolved. I suggest that the JEP authors make any final updates to the language, or resolve any questions you think need resolving, and then mark the JEP as "ready for review". This will trigger the voting phase, at which point the language should not substantially change.

NN-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,127 @@
...
+- As a user, I want extensions that need a left/right/bottom/status area to work in Notebook, but I want those areas to be entirely hidden by default, so I can have the document-oriented notebook experience that I am used to.
+- As a user, I want to know all of the great new things I am getting with Notebook v7, so I can make an informed decision to upgrade.
+- As a user, I want the most common notebook extensions to work in both the notebook and lab interfaces, so I can move back and forth without missing functionality (see below).
+-   As a user, I want this user experience to work seamlessly with the [Jupyter Desktop](https://blog.jupyter.org/jupyterlab-desktop-app-now-available-b8b661b17e9a)
+    - As a user, I can double-click on an `.ipynb` file on my local computer (Mac, Win, Linux) to open the notebook as a single document which I can edit, run, etc.
+    - As a user, I want to be able to go from the document-centric experience to the full JupyterLab Desktop application, so I can get a more IDE-centric experience, including file management, terminals, and more.
+    - As a user, I can open notebooks via the command-line or open JupyterLab in a particular directory.
+
```
ellisonbg:
```suggestion
- As a JupyterLab extension author, I want my extension to work with both Notebook 7 and JupyterLab 4, without maintaining two code bases or releasing two packages.

```

fperez: 👍

ellisonbg:
I have made a suggestion with one additional users story that is focused on the extension author persona and their experience of supporting extensions in both Notebook 7 and Lab 4.

fperez:
@choldgraf I made a tiny metadata update and now marked it officially as 79 in the folder name. I think it's ready to go...

Oh wait! I just saw @ellisonbg's update - I agree with it but can't seem to push it from the UI. Weird...

fperez:
I guess I don't have write permissions to make the commit. Odd... I'll do it manually on the branch now then.

fperez:
@ellisonbg do you have any others? Else we can move it out of draft status...

NN-notebook-v7/notebook-v7.md
jtpio: Looks good, thanks Brian :+1:

ellisonbg:
> @ellisonbg do you have any others? Else we can move it out of draft status...

Nope, I think that is it, I am fine with us moving it out of draft.

fperez:
OK 🚀! @choldgraf how do I turn off the draft flag on this? I can't seem to find the UI element you'd showed me earlier, not sure if it's a repo permissions issue...

I've removed "DRAFT" from the title though - this is **ready for review**.

NN-notebook-v7/notebook-v7.md
choldgraf: I think this is a good constraint to design for!

choldgraf:
It should be just above the comment box: 

image

fperez:
Ah, thanks! I was looking for it in up above and missed it below. Duh.

fperez:
Huge thanks to everyone who made it possible to get us here. I'll post now on discourse and send a quick email to the SC (which right now we need for voting) so this moves forward promptly.

choldgraf:
Hey all - I believe that the Review Team for this JEP is the @jupyter/steeringcouncil - I have added a voting list to the top of this document (I copy/pasted the same list of names from #72).

### Instructions

For the @jupyter/steeringcouncil - please check the box next to your name according to how you'd like to vote on this JEP. If you wish to propose minor changes (e.g. small language or typos), that is OK. If you wish to propose changes that are significant, mark your vote as "no", as we cannot change the JEP language significantly without starting the process again.

Carreau (approved):

SylvainCorlay (approved):

minrk (approved):

choldgraf:
### Process update: Try to get comments in by 2021-12-13

I wanted to provide a process update to this JEP. The [JEP guidelines suggest we use a 10 day minimum for this Final Comment Period](https://jupyter.org/enhancement-proposals/jupyter-enhancement-proposal-guidelines/jupyter-enhancement-proposal-guidelines.html#phase-2-request-for-comments-for-the-jep). This date would be roughly the 13th.

In addition, I have run into a few confusion points for myself in shepherding this PR, and have opened an issue to document some of these so that we can do process improvements in the future (#83)

79-notebook-v7/notebook-v7.md
```diff
@@ -0,0 +1,127 @@
...
+
+There will always be tradeoffs between these two options. To help us make these tradeoffs, we propose to look at the following questions:
+
+* Is Option A or Option B be more accessible?
+* If we were to design this from scratch today, would our design work and look more like Option A or Option B?
+* Is Option A or Option B more maintainable?
+* Is Option A or Option B be more secure?
+* Will Option A confuse or cause friction to Jupyter users who work in both Notebook V7 and JupyterLab v4?
+* What is best for future Jupyter users who have never used either any version of Notebook or JupyterLab?
+* Have we done everything possible to minimize the experience gap for users of Notebook v6 (while recognizing, given other constraints, that we can't always eliminate it completely)?
```
Carreau, line 110: I had an in person discussion recently that pushed me to comment on this section. 

I'm in no way asking for a change, but I am going to write something I've said in the original governance call where this was started, and was unwilling to say/write again. Especially it was pointed out to me that others may feel similarly might not feel like expressing themselves.

I am aware that especially for this project it is better to all push in the same direction, and I hope that at least posting those thoughts here will at least reassure users that those concerns were expressed, heard, even if no change in current document are acted.

--- 

I am afraid this is the same framework that was used for JupyterLab, and for me this is not biased enough toward the "classic" notebook behavior. And I do believe this should be biased toward v6 users. 
Our goal here is to provide a palatable transition to current classic notebook users. This framework and way of phrasing decisions targeted toward providing something "better" for end user (for whatever metric of better we  choose), which tilts highly toward jupyterLab.

IMHO the framework would be better suited for a "version 7.5" than "version 7.0".
Especially since we don't have infinite manpower, and that we don't want to rehash all and every discussion I would have other concerns:

 * Is Option A or Option B be more accessible?
     * Is option A less accessible than current notebook classic v6 should be enough of a question.
 * If we were to design this from scratch today, would our design work and look more like Option A or Option B?
     *  irrelevant, we know what  the answer it's : the newer is jupyterlab so the response to this questions will always be B . The goal is to get users from classic to the new version.
 * Is Option A or Option B more maintainable?
     *  This is also a non question, the answer will always be B: jupyterlab. The real question is always how much more maintenance burden we are willing to take to look more like classic.
* Will Option A confuse or cause friction to Jupyter users who work in both Notebook V7 and JupyterLab v4?
  * Again wrong questions IMHO, and answer will always be B. We are not trying to cater to users using both. The question should be "is it worse than for users of v6 and Lab". Remember most users we are doing this for do do not use lab. It those users we are trying to please.
* What is best for future Jupyter users who have never used either any version of Notebook or JupyterLab?
  *  This is again broadening the scope and is the same questions that were asked and led to lab, and also not the users we should be catering to for 7.0

So while I as a person agree with the principle in the absolute and hope they will be reached in the future, I think in this context they are misguided, and would suggest caution of scope creep and repeating previous mistakes. 

To TL:DR, we're not trying to write directly python 3.6 after 2.7, we are trying to write a 3.0 that sucks less for 2.7 users, like https://www.python.org/dev/peps/pep-0414/ (re-adding u-prefix to strings). Or for Physicist we are trying to have a quasi-static transformation, not a lower energy final state.

I will of course be a strong supporter of any decision that is taken, and trust the people more engaged in this than I. I also completely believe that even with the above framework the people involved will do the right decisions, but I'm concerned that it leaves the door open for interpretation and bikeshedding of every decision, and expands the scope of works far beyond what is necessary. I also believe the a clear, narrow and non-ambiguous scope of work make it much easier for drive-by and new contributors to participate as there is no frictions due to decision making. 

--- 

So here is what I was refraining of posting, apologies as this is during the end of the comment period,I again do not ask for changes, and I am looking forward to notebook 7 !

Carreau (approved):

79-notebook-v7/notebook-v7.md
choldgraf: @Carreau to make sure I understand, it sounds like you are suggesting that during the implementation of this JEP, a good rubric is:

- For the purposes of Notebook V7, we should shoot for the minimal possible Notebook UI that builds on JupyterLab components, and that replicates Notebook V6 functionality. This is Notebook V7.0
- Adding extra functionality beyond replicating Notebook V6 is a "nice to have", but should be secondary steps after we satisfy the step above. Better to do those iteratively, and imagine those as Notebook V7.1/2/3 etc.
- Notebook V7.8 will redefine the `.ipynb` file format to be built entirely with walrus operators `:=` (I kid 🙂)

Is that right?

ellisonbg: I think the questions in this section are less about what features/functionality Notebook V7 should have and more about how we should offer the features that are common between Notebook V7 and Lab 4 in terms of visual design, interaction design, behavior, etc.

SylvainCorlay:
Quick update on the front of extension support in notebook v7:

 - **RISE** is currently being ported to the new extension system by @fcollonval, with help from @damianavila and @parmentelat. Cf https://github.com/damianavila/RISE/pull/605.
 - The QuantStack team just had a quote approved to port the UI of **nbgrader** to the JupyterLab extension system. Work on nbgrader should start in January! 🎉

afshin:
> The QuantStack team just had a quote approved to port the UI of nbgrader to the JupyterLab extension system. Work on nbgrader should start in January! 🎉

That's awesome news!

cc @jhamrick

79-notebook-v7/notebook-v7.md
Carreau: To Chris; Yes. Except I think the walrus operator should actually be a match statement.

To Brian, yes, and that is what I disagree with. The problem with framing that as "between V7 and JupyterLab" is that if frames V7 as a JupyterLab alternative, and that is not the reason we are doing this Jep:

> This PR proposes a path forward for the user experience of the jupyter notebook application

 The comparison should be V6, always. The fact that we are using JupyterLab components is an implementation detail – and important one for sure, but Lab behavior should not be the goal of the UI/UX. And I really really would like to avoid any mention that compare V7 to JupyterLab.

If I had to frame it in a different way, for me:
  - for v7.0: Any user facing difference in behavior from V6 is bug. The question is are we ok to live with it because the cost is too high.

There is nowhere in the above mention of Lab. Nor should there be.

Now if you want to tell me "for 7.3 we actually want to remove the "l" shortcut to toggle line number per cell as it's more consitent with Lab to always toggle all cells" i'm fine with that. But I don't think it is  for 7.0.

choldgraf: The way I think about it is that when Notebook v7.0 is released, there are two voices that will be in a user's head:

1. "I hope I can still do XXX - how do I do it?" (aka, we expect people want the same workflow in V7, and should minimize the disruption of the change)
2. "I wonder what else I can do now that things are different!" (aka, people may want to extend their workflows into new directions with new functionality to explore what is possible)

I think @Carreau is suggesting (and I think I agree w/ this) that for an initial 7.0 release, it is more important to minimize `#1` than it is to maximize `#2`. Finding new workflows will be useful for people, but it will happen more organically and over time. People are less likely to explore what the tools can do for them in the future if they immediately feel like their core workflow is disrupted. Moreover, people that strongly benefit from the workflows in JupyterLab are more likely to have already switched. For this reason, it's better to focus on "reduce the transition shock from V6 to V7" in this initial step.

SylvainCorlay: > Now if you want to tell me "for 7.3 we actually want to remove the "l" shortcut to toggle line number per cell as it's more consitent with Lab to always toggle all cells" i'm fine with that. But I don't think it is for 7.0.

The effort that is currently ongoing in Retrolab is to get as close as possible to the notebook v6 user experience. Although there will definitely be *some* differences. The model that is proposed here is that any gap that remains between v6 and v7 should be small enough that users don't have a problem making the transition.

gutow (commented):
As someone who develops specialty use tools/extensions I am worried that something important to my end users is missing in the user stories. First the background and then what I think my users would want:
**Background**
I have been investigating what some of the tools I have developed would look like to my end users if I convert them. Currently, my end users can install my tools simply via pip in the environment the jupyter notebook runs from. It appears that will no longer be the case because some of my tools use custom javascript, not just python. If I am understanding things properly the users will also have to have an NPM environment installed, so that the `jupyter lab build` command can incorporate the javascript. I base this on the Lab compatible version of the [snippets](https://github.com/QuantStack/jupyterlab-snippets) extension, because it is close to one of the things I do (providing custom menus that depend on what is being run in a particular notebook).
**What my users want***
They want a single step or command for installing extensions. 

For many of my end users even learning to use `pip` or (`%pip`) is a big deal, because they do it so rarely. This means they might not successfully be able to migrate to the Retrolab based notebook alternative. I believe that thinking of people like me who write some code that is more than just short scripts run in Jupyter as the end user is the wrong way to think about this. I think the real end-user you need to target are people like my colleagues in academia who want to install tools that provide menus, GUIs, etc. to facilitate their students and them doing exercises within the Jupyter environment, without having to worry about much about how to set it up. My colleagues will not switch to the newer system if each extension/adjustment requires more effort than it used to.

krassowski:
I fully agree that having the UI for extension manager that allows to install prebuilt extensions equivalents for each of the extensions that were available from notebook contrib should be part of the plan.

Currently the Extension Manager in JupyterLab installs source extensions which leads to a lot of issues as discussed in https://github.com/jupyterlab/jupyterlab/issues/10554. The current plan is to remove the source extensions searching from the core for 4.0 and to allow extensions to provide lists of extensions that can be installed (as well as to let them handle the installation itself).

While I fully agree that we should switch to a modular approach and drop source extensions (in favour of prebuilt extensions which are easier to install), I believe that ideally we should also include a "good enough" provider of extensions that handles the most common use cases of installation in a Python virtual environment and in a conda environment.

During the recent JupyterLab meetings there was a consensus that this is a hard problem (conda vs pip vs non-Python packages, querying PyPI API to create the list of extensions, etc), but I think it is manageable.

bollwyvl:
>  users will also have to have an NPM environment installed... snippets

Ah, but the fine print about `jupyter lab build`: _If you use JupyterLab 2.x,_

It's probably worth re-iterating: 

**No end user feature of Notebook 7 will require `nodejs` or any other non-`pip install`able package**

I think by the time _Notebook 7_ lands (some time after _Lab 4_), all important extensions where the developers intended to supporting them _at all_ will already offer one-step, `pip install`able pre-built extensions. 

> They want a single step or command for installing extensions.

As with `nbextensions` in _Notebook <7_, end users will be able to `pip install` or `conda install` new web UI features, requiring a _browser reload_ to use new UI features. 

There were certain ways this _could_ be made to work well in _Notebook <7_ without a page reload (e.g. widgets), but we could never offer sound compatibility guarantees for everything: just slamming some JS out of a kernel into an output and expecting a new top-level menu item to show up is _not_ a capability we can maintainably offer in _Notebook 7_.

**Server** extensions (e.g. new REST APIs) will continue to require a restart of the `notebook` server... but this has always been a system property of the notebook server. This could _theoretically_ be fixed at an architectural level... but has the implicit potential, especially in a managed/Hub setting, of leaving a user with a completely wedged, unstarteable, uninspectable, remote web server, which has always prevented anyone from pushing too much harder on it.

> we should also include a "good enough" provider of extensions 

Yeah, it would be nice... but as there never was an in-app "Extension Store," in _Notebook <7_, I _don't_ think, it's in-scope for _this_ JEP.

bollwyvl:
> Are there clear examples I can follow that do this using menus that inject code into selected cells (the key thing I do)?

The snippets example you linked to has most of the pieces. Similarly, the [dask-labextension](https://github.com/dask/dask-labextension) shows some approaches for working with the context menu.

>  trouble getting some of them running at all

without any more info, it's hard to say... but the venue for further discussion is likely issues on the extensions themselves, where it could prove out _no, this developer does not intend to support this extension_ (which is of course their choice).

> ways to control what menus and buttons show via python

At present, there isn't a way for (any) kernel to know some "client name" against which they are running (it could be multiple, simultaneously). Again, this could be made more lax, but brings a host of other deep issues, and is outside the scope of this JEP. But basically, kernel-emitted JS code that only works in one client (e.f. _Notebook <7_) probably already didn't work in any other first-party (lab, nbconvert) or third-party client (nteract, vscode, spyder, colab). Part of the maintainability issue we are addressing here is precisely this... perhaps widgets or mimetypes for these kinds of things are the way to move forward. had a comment about that earlier, but it appears to be gone.

choldgraf:
Hey all - we're now outside the "final comment period" minimum window. I'll note that several @jupyter/steeringcouncil members haven't given their thoughts or a vote yet. Could somebody send another message to that group (or tell me how I can best get in touch with the group) to let them know we'd like final thoughts and comments before moving forward?

gutow:
> > ways to control what menus and buttons show via python
> 
> At present, there isn't a way for (any) kernel to know some "client name" against which they are running (it could be multiple, simultaneously). Again, this could be made more lax, but brings a host of other deep issues, and is outside the scope of this JEP. But basically, kernel-emitted JS code that only works in one client (e.f. _Notebook <7_) probably already didn't work in any other first-party (lab, nbconvert) or third-party client (nteract, vscode, spyder, colab). Part of the maintainability issue we are addressing here is precisely this... perhaps widgets or mimetypes for these kinds of things are the way to move forward. had a comment about that earlier, but it appears to be gone.

I understand this to mean that I would have to have my students launch different versions of JLab depending on their use case. At that point, might I be better off writing dedicated applications using QT or wx? However, I do not really want to have to duplicate the very useful simple notebook interface. The other alternative is to have a single heavy weight version of JLab with lots of menus, buttons, etc. that they don't need most of the time. In that case, I would be producing a specialized version of JLab that would have many of the problems for my use case that this JEP is trying to find a path to avoid.

I am also confused by "there isn't a way for (any) kernel to know some "client name" against which they are running." One should be able to look at either existing DOM elements or interfaces to determine what is appropriate. If the necessary information is completely hidden or obfuscated, then I believe you have lost one of the key advantages of the classic notebook.

I have no objection to having to maintain two different javascript codes during the transition from the old notebook to a JLab based one. There just needs to be a clear way to do that. I am struggling with getting what were relatively simple things to do in the old notebook working in JLab. I believe it is primarily due to the opacity of the API and lack of good examples. When I say opacity, I mean it is often hard to tell without stepping back through hundreds of lines of code exactly what options, variables, etc to pass to the high level API calls. I think a concerted effort on documentation and examples would go a long way to improving this situation.

I am strongly in favor of finding a path forward. I am just worried that as mentioned by @choldgraf and @Carreau that too much of the capabilities available when working within the Classic Notebook will be lost for many of us to effectively make the transition. I personally would like to see more language concerning support for more of the on-the-fly flexibility of the classic notebook. I do understand that that will be less compatible with things outside of JLab and has some security issues. From my perspective the ease of producing something I can use for 5 - 10 years with less development overhead is more important.

ivanov:
> Could somebody send another message to that group (or tell me how I can best get in touch with the group) to let them know we'd like final thoughts and comments before moving forward?

I just sent a nudge to the group and the individuals who haven't registered their decisions here, @choldgraf

rgbkrk:
Thanks for the nudge.

rgbkrk (approved):
Let's move the notebook forward!

damianavila (approved):

Zsailer:
It's worth noting here that there's a live, weekly discussion happening around this JEP and the pending work in the weekly Notebook meeting. Here is a record of the notes from previous weeks: https://github.com/jupyterlab/team-compass/issues/133. 

As always, these meetings are open to anyone and everyone. Feel free to join the conversation. It's always a hoot! 😄

jasongrout (approved):

jasongrout:
Thank you very much for moving this forward!

choldgraf:
Hey all - we have had a few weeks of time for comments and conversations, and we've gone through many rounds of discussion and edits (both in this PR and in the linked issues). Here's a tally of the @jupyter/steeringcouncil votes so far:

- Yes: 10
- No: 1
- Abstain: 2
- No response: 4

### Proposal

As we have a strong majority of the steering council voting in favor of this, and have had a lot of productive conversations thus far in other issues and here, I propose the following steps:

- Leave this PR open for final comments until **Tuesday the 21st.**
- Unless a major new piece of information arises that has not already been discussed, we merge this PR at EOD that Tuesday.
- I will update the top comment with a summary of the discussion here (with a focus on major concerns people have that could impact implementation decisions), so that it is easier to refer to in the future.
- We consider this JEP **accepted**

ivanov:
I was asked privately why I voted against this proposal, and my reply was deemed to contain aspects of the issues at hand that were not covered in this or some of the other threads around the proposal, so I am reposting the bulk of it here.

This isn't exhaustive but here goes. I don't agree with changing the entire project and keeping the same name, which is what notebook7 will be. I can't think of quite an equivalent thing here, but it would be like installing notepad, but getting Word that's themed to look like notepad. I guess the positive example here was Mozilla Application Suite / SeaMonkey, the re-written browser component of which is where Firefox came from, but when Firefox came out, it was a new thing. People didn't get it as an update to SeaMonkey. It would have been more responsible to make a new name for this and let people migrate to it. I also do not think JupyterLab as the code base is the thing to get behind. Additionally, I do not think there is any urgency to considering this change now - we've been in limbo on governance and decision making for the entire project for quite some time, and yet this was pushed through without wrapping up the governance refactor before prioritizing other business. I understand that it is difficult to push forward on any front with so much up in the air, but I would have preferred for us to wrap up to the new decision making process and navigated the decisions of this JEP in the new governance configuration.

Carreau:
> I don't agree with changing the entire project and keeping the same name, which is what notebook7 will be. I can't think of quite an equivalent thing here, but it would be like installing notepad, but getting Word that's themed to look like notepad.

While I understand this point I will point you toward GCC/EGCS. I think that if the replacement is good/close enough it is worth it to have a more cohesive community. But again this is not trying to convince you, just maybe giving you a historical example. This is also why I've strongly pushed toward not phrasing comparison WRT jupyterLab, but notebook v6. 

I also do see the reason for the same name with respect to user muscle memory being to `pip install notebook`

blink1073:
The voting window has completed, merging as accepted, thanks everyone!

choldgraf:
Just wanted to say thanks @ivanov for providing your thoughts here - I think it complements @Carreau's thoughts as well, that Notebook v7 should be thought of as "an evolution of Notebook v6" rather than "an adaptation of JupyterLab". And thanks @blink1073 for merging it in 👍 

I will update the top comment with some of the major conversation points here, so that is it easier to discover in the future if people refer back to this PR thread.

matthew-brett:
This is really just to follow up on the linked issue at https://github.com/jupyter/notebook/issues/6210 - or at least - to ask where to follow up about that. 

That issue was raising the question as to whether JupyterLab had fallen between two stools, these stools being:

* Simplicity of use and extension by not-professional programmers (well covered by (now) `nbclassic`) and
* Power for customization for professional programmers.

And further, whether falling between these stools might make it easier for e.g VSCode to outcompete JupyterLab.

I understood that Notebook version 7 was largely designed to look more like the user experience of (now) `nbclassic`, and there was also some effort underway to make it easier to write JupyterLab extensions.

My questions are - how successful were those two efforts in a) bringing people across from `nbclassic` to JupyterLab, and b) has there been any uptick in i) use of JupyterLab compared to say VSCode for editing notebooks and / or ii) development of JupyterLab extensions by not-professional programmers?

jasongrout:
> I'm not sure if the issue should be on this repo, but I don't have a much better place to suggest. If it's actual Data, maybe the Executive council repo, otherwise maybe zulip and/or jupyter discourse forum.

As both of these packages are in the frontends subproject, I suggest the appropriate place for this conversation is in the frontends subproject, for example the [team compass](https://github.com/jupyterlab/frontends-team-compass), discourse, zulip, or the weekly dev meeting that includes the frontends projects (in reverse order of bandwidth/immediacy of the discussion).

matthew-brett:
Thanks for the data, that's interesting.

I posted here because I was worried about the results of the overall strategy to win users for JupyterLab from the classic notebook, and from VSCode.  There were signs, at the time, that uptake seemed to be relatively slow, especially given the age of the Classic notebook.  So I wanted to keep the follow up with the previous discussion, to make clearer the link between the previous concerns, the proposed solution, and the predicted outcome.

At the time, I think it's fair to say that the team's response was more or less - we should continue and speed up the then-current strategy of simulating the Classic experience with JupyterLab 7, and improving the developer tools and docs for extensions.   I was (others were) worried that a bolder strategy might be needed.

So I was hoping that the team had done some further evaluation as to how the proposed strategy had worked out- in order to reassess strategy in view of the results.

PR #80, opened 2021-12-02
Convert build to a Sphinx site and minor updates to docs

choldgraf:
This PR does a few main things:

- Updates the repository to be a Sphinx site instead of a Jupyter Book. This gives us a bit more control over things and lets us preview the documentation on ReadTheDocs. It retains the same structure and look as if we used Jupyter Book though.
- Adds links to a few JEPs that were not added to the table of contents when they were accepted.
- Adds numbering to these JEPs to make them more in-line with our other JEPs
- Adds the current active JEP in #79 to our README
- Adds instructions to our README for how to build the docs locally
- Adds a `noxfile.py` so that we can automate building docs via `nox`
- Removes the CircleCI build, since we are now building the docs with ReadTheDocs previews

We can now preview the built docs in PRs, like so: https://jupyter-enhancement-proposals--80.org.readthedocs.build/en/80/README.html

PR #81, opened 2021-12-02
Remove GH artifact action and update workflow for gh-pages

choldgraf:
One quick follow-up! This removes a no-longer-needed artifact generation action, and also updates our GH Pages action to use the same workflow we recommend for building locally. It also updates the README to explain this a bit!

(sorry for not catching this in the last PR)


choldgraf:
I will self-merge this one, as it is a minor follow-up to #80 !

PR #82, opened 2021-12-03
Support kernel_info request on the control channel

JohanMabille:
In this JEP we propose the `kernel_info` request can be sent on both the shell and the control channels, to fix issues like [this](https://github.com/jupyter-server/jupyter_server/issues/622) and [this](https://github.com/jupyterlab/jupyterlab/issues/10174).

cc @SylvainCorlay and @minrk who I have talked to about this JEP.

- @echarles 
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @fcollonval 
  - [x] Yes
  - [ ] No 
  - [ ] Abstain
- @gabalafou
  - [ ] Yes
  - [ ] No
  - [x] Abstain 
- @ibdafna 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @ivanov
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @JohanMabille 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @marthacryan
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @minrk
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @rpwagner 
  - [ ] Yes
  - [ ] No
  - [ ] Abstain 
- @SylvainCorlay
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @Zsailer 
  - [x] Yes
  - [ ] No 
  - [ ] Abstain

### DIFF
--- a/80-kernel-info/kernel-info.md
+++ b/80-kernel-info/kernel-info.md
@@ -1,4 +1,4 @@
-# Send kernel\_info request on the control channel
+# Support kernel\_info request on the control channel
 
 ## Problem
 
--- a/80-kernel-info/kernel-info.md
+++ b/80-kernel-info/kernel-info.md
@@ -16,7 +16,7 @@ We propose to state in the Jupyter Messaging Protocol that the `kernel_info` req
 
 This JEP impacts kernels since it requires them to support receiving 'kernel\_info\_request' on the control channel in addition to receiving them on the shell channel.
 
-It also has an impact on  the Jupyter Server, which must be updated accordingly.
+It also has an impact on the Jupyter Server. For example, the reference implementation of Jupyter Server will attempt to send a a `kernel_info` request on both channels and listen for a response from _either_ channel. Any response informs the UI that the kernel is connected. 
 
 ## Relevant Resources (GitHub repositories, Issues, PRs)
 
--- a/80-kernel-info/kernel-info.md
+++ b/80-kernel-info/kernel-info.md
@@ -14,7 +14,7 @@ We propose to state in the Jupyter Messaging Protocol that the `kernel_info` req
 
 ### Impact on existing implementations
 
-This JEP impacts kernels since it requires them to support receiving 'kernel\_info\_request' on the dontrol channel in addition to receiving them on the shell channel.
+This JEP impacts kernels since it requires them to support receiving 'kernel\_info\_request' on the control channel in addition to receiving them on the shell channel.
 
 It also has an impact on  the Jupyter Server, which must be updated accordingly.
 
--- a/80-kernel-info/kernel-info.md
+++ b/80-kernel-info/kernel-info.md
@@ -1,0 +1,40 @@
+# Send kernel\_info request on the control channel
+
+## Problem
+
+When connecting a new websocket to an existing kernel via the Jupyter server, if the kernel execution is
+stopped on a breakpoint, the messages sent over the websocket get no reply. This is because when
+establibshing a new websocket connection, the Jupyter server will send `kernel_info` requests to the kernel
+and will prevent sending any other request until it receives a `kernel_info` reply. Since the `kernel_info`
+request is sent on the shell channel and the kernel executino is stopped, it cannot reply to that request.
+
+## Proposed enhancement
+
+We propose to state in the Jupyter Messaging Protocol that the `kernel_info` request muter st be sent on the
+control channel exclusively.
+
+### Impact on existing implementations
+
+There should not be any impact on the existing kernels, since the current specification states that any message
+that can be sent on the shell channel could be sent on the control channel.
+
+The only impact of this change (beyond the protocol itself) is on the Jupyter Server, which must be updated
+accordingly.
+
+## Relevant Resources (GitHub repositories, Issues, PRs)
+
+### GitHub repositories
+
+- [Jupyter server](https://github.com/jupyter-server/jupyter_server): the backend to Jupyter web applications
+- [Jupyter client](https://github.com/jupyter/jupyter_client):  Jupyter protocol client APIs
+- [JupyterLab](https://github.com/jupyterlab/jupyterlab): JupyterLab computational environment
+
+### GitHub Issues
+
+- New websocket unable to communicate with kernel paused in debugging [#622](https://github.com/jupyter-server/jupyter_server/issues/622)
+- [Debugger] Active sessions cause the kernel to deadlock on page refresh [#10174](https://github.com/jupyterlab/jupyterlab/issues/10174)
+
+### GitHub Pull Requests
+
+Nudge kernel with info request until we receive IOPub messages [#361](https://github.com/jupyter-server/jupyter_server/pull/361)
+
--- a/80-kernel-info/kernel-info.md
+++ b/80-kernel-info/kernel-info.md
@@ -3,7 +3,7 @@
 ## Problem
 
 When connecting a new websocket to an existing kernel via the Jupyter server, if the kernel execution is
-stopped on a breakpoint, the messages sent over the websocket get no reply. This is because when
+busy (e.g. long running cell) or stopped (e.g. on a breakpoint), the messages sent over the websocket get no reply. This is because when
 establishing a new websocket connection, the Jupyter server will send `kernel_info` requests to the kernel
 and will prevent sending any other request until it receives a `kernel_info` reply. Since the `kernel_info`
 request is sent on the shell channel and the kernel execution is stopped, it cannot reply to that request.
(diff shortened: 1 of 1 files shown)


80-kernel-info/kernel-info.md
```diff
@@ -0,0 +1,40 @@
+# Send kernel\_info request on the control channel
+
+## Problem
+
+When connecting a new websocket to an existing kernel via the Jupyter server, if the kernel execution is
+stopped on a breakpoint, the messages sent over the websocket get no reply. This is because when
+establibshing a new websocket connection, the Jupyter server will send `kernel_info` requests to the kernel
+and will prevent sending any other request until it receives a `kernel_info` reply. Since the `kernel_info`
+request is sent on the shell channel and the kernel executino is stopped, it cannot reply to that request.
```
davidbrochart, line 9:
```suggestion
request is sent on the shell channel and the kernel execution is stopped, it cannot reply to that request.
```

80-kernel-info/kernel-info.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Proposed enhancement
+
+We propose to state in the Jupyter Messaging Protocol that the `kernel_info` request muter st be sent on the
```
davidbrochart, line 13:
```suggestion
We propose to state in the Jupyter Messaging Protocol that the `kernel_info` request must be sent on the
```

80-kernel-info/kernel-info.md
```diff
@@ -0,0 +1,40 @@
...
+that can be sent on the shell channel could be sent on the control channel.
+
+The only impact of this change (beyond the protocol itself) is on the Jupyter Server, which must be updated
+accordingly.
+
+## Relevant Resources (GitHub repositories, Issues, PRs)
+
+### GitHub repositories
+
+- [Jupyter server](https://github.com/jupyter-server/jupyter_server): the backend to Jupyter web applications
```
davidbrochart, line 28: Maybe add jupyverse?
```suggestion
- [Jupyter server](https://github.com/jupyter-server/jupyter_server): the official backend to Jupyter web applications
- [Jupyverse](https://github.com/jupyter-server/jupyverse): a FastAPI-based backend to Jupyter web applications
```

ivanov:
From the description of this proposal, it seems like the problem at hand can be solved at the Jupyter Server level, without introducing backwards incompatible protocol changes. Can jupyter server be changed to issue its initial `kernel_info_request` on the control channel, and also not assume it has to receive a reply to that message before deeming a connection as established? 

It makes sense to _encourage_ clients to issue `kernel_info` requests over control channel, but dropping support for it on the shell channel would be a major protocol version change.

JohanMabille:
@ivanov I agree that a backward incompatible change is not ideal. We could do what we did with `kernel_interrupt` and `kernel_shutdown` requests: we move the description of the message to the `Messages on the control channel` section of the protocol, and mark it as deprecated on the shell channel. We would have nothing to change in the implementation of Jupyter Server for now, since it sends `kernel_info` on both channels thanks to the last fix by @Zsailer. If you agree with that, I will update the JEP accordingly.

ivanov:
@JohanMabille I'm still not sure that we want to necessarily deprecate the message pair on the shell channel. I understand that it's easier for implementors if a given message should only appear on one channel. However, one of the affordances with the shell channel approach is the current pattern of clients issuing a `kernel_info_request` on connect, and indicating that the connection is not complete and that the kernel is busy or otherwise not yet known to be available *until* it gets back a reply. So it serves as a queued no-op with respect to kernel state, and does not require the client to know the language of the kernel. 

Specific example would be similar to the debugging example in the JEP but suppose that instead of debugging, you have a long running computation (like `sleep 100`) that is running when a new client connects. If the new client makes a kernel info request on the control channel, and the kernel handles that request, the client still does now know whether the kernel is in busy or idle state until it makes an execute request. And even after making an execute request, client has no way of indicating if the pending execution is for the new code, or for the completion of previously running code, followed by the new code.

One possibility could be to add a kernel status message to the body of a `kernel_info_reply` content, to resolve this ambiguity for the client in the absence of kernel status messages.

SylvainCorlay:
On the core of this proposal, we clearly need to require kernel info to be supported on the control channel, for situations such as connecting to a kernel that is currently blocked on a breakpoint or currently executing a long-running cell. In my opinion, this should be resolved at the protocol level because it also applies to cases where there isn't a server (connecting to an existing kernel).

On the other point brought up by @ivanov on the semantics of kernel idle and busy messages: We would like to make some proposals at a later stage on this subject. My take on this is that idle and busy messages should eventually only concern the state of the Shell channel (it is not of interest to kernel end-users to know whether the control channel is processing messages. Control channel messages ought to be processed fast in a non-blocking way as much as possible).
Kernels such as [akernel](https://github.com/davidbrochart/akernel) which can process several execution requests concurrently also challenge the current model.

80-kernel-info/kernel-info.md
```diff
@@ -0,0 +1,41 @@
+# Send kernel\_info request on the control channel
+
+## Problem
+
+When connecting a new websocket to an existing kernel via the Jupyter server, if the kernel execution is
+stopped on a breakpoint, the messages sent over the websocket get no reply. This is because when
+establibshing a new websocket connection, the Jupyter server will send `kernel_info` requests to the kernel
```
gabalafou, line 7:
```suggestion
establishing a new websocket connection, the Jupyter server will send `kernel_info` requests to the kernel
```

JohanMabille:
I rewrote the proposal to state that the `kernel_info` request should be supported on both Shell and Control.

minrk (approved):

80-kernel-info/kernel-info.md
```diff
@@ -0,0 +1,37 @@
...
+and will prevent sending any other request until it receives a `kernel_info` reply. Since the `kernel_info`
+request is sent on the shell channel and the kernel execution is stopped, it cannot reply to that request.
+
+## Proposed enhancement
+
+We propose to state in the Jupyter Messaging Protocol that the `kernel_info` request can be sent on both the shell and the control channels. Although both channels supports the `kernel_info` message, clients are encouraged to send it on the control channel as it will always be able to handle it, while the shell channel may be stopped.
+
+### Impact on existing implementations
+
+This JEP impacts kernels since it requires them to support receiving 'kernel\_info\_request' on the dontrol channel in addition to receiving them on the shell channel.
```
gabalafou, line 17:
```suggestion
This JEP impacts kernels since it requires them to support receiving 'kernel\_info\_request' on the control channel in addition to receiving them on the shell channel.
```

80-kernel-info/kernel-info.md
```diff
@@ -0,0 +1,37 @@
+# Send kernel\_info request on the control channel
```
gabalafou, line 1:
```suggestion
# Support kernel\_info request on the control channel
```

80-kernel-info/kernel-info.md
```diff
@@ -0,0 +1,37 @@
+# Support kernel\_info request on the control channel
+
+## Problem
+
+When connecting a new websocket to an existing kernel via the Jupyter server, if the kernel execution is
+stopped on a breakpoint, the messages sent over the websocket get no reply. This is because when
```
Zsailer, line 6:
```suggestion
When connecting a new websocket to an existing kernel via the Jupyter server, if the kernel execution is
busy (e.g. long running cell) or stopped (e.g. on a breakpoint), the messages sent over the websocket get no reply. This is because when
```

80-kernel-info/kernel-info.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Proposed enhancement
+
+We propose to state in the Jupyter Messaging Protocol that the `kernel_info` request can be sent on both the shell and the control channels. Although both channels supports the `kernel_info` message, clients are encouraged to send it on the control channel as it will always be able to handle it, while the shell channel may be stopped.
+
+### Impact on existing implementations
+
+This JEP impacts kernels since it requires them to support receiving 'kernel\_info\_request' on the control channel in addition to receiving them on the shell channel.
+
+It also has an impact on  the Jupyter Server, which must be updated accordingly.
```
Zsailer, line 19:
```suggestion
It also has an impact on the Jupyter Server. For example, the reference implementation of Jupyter Server will attempt to send a a `kernel_info` request on both channels and listen for a response from _either_ channel. Any response informs the UI that the kernel is connected. 
```

JohanMabille:
7 yes, 1 abstention, and 2 non votes, let's merge!

PR #93, opened 2023-02-20
JEP for copy-to-globals support in debugger_info

brichet:
Introducing a new field to the kernel debugger_info response. 
This new field will inform the UI that the debugger supports the `copyToGlobals` request.

Related to https://github.com/jupyterlab/jupyterlab/pull/13476#discussion_r1111061085, which introduces a new feature in the debugger. It would be useful to know if this feature is supported by the kernel from UI side.

cc. @JohanMabille @krassowski 

- @echarles 
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @fcollonval 
  - [x] Yes
  - [ ] No 
  - [ ] Abstain
- @gabalafou
  - [ ] Yes
  - [ ] No
  - [x] Abstain 
- @ibdafna 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @ivanov
  - [ ] Yes
  - [ ] No
  - [x] Abstain
- @JohanMabille 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @marthacryan
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @minrk
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @rpwagner 
  - [ ] Yes
  - [ ] No
  - [x] Abstain 
- @SylvainCorlay
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @Zsailer 
  - [x] Yes
  - [ ] No 
  - [ ] Abstain

### DIFF
--- a/debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
+++ b/debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
@@ -23,3 +23,47 @@ UI if this is supported by the kernel before displaying the corresponding menu e
 
 We propose to add a new `copyToGlobals` boolean field to the `debugger_info`
 response which will inform the UI that this request is supported.
+
+## Reference-level explanation
+
+This boolean flag should be included in the `debugger_info` response from the kernel
+which supports the feature. It is optional, assuming that its absence is understood
+as `false` from the client perspective.
+
+If the feature is supported, the kernel must provide a function for copying a variable
+from a local scope to the global scope.
+The communication between the UI and the kernel (request - response) will have the
+structures described at
+https://jupyter-client.readthedocs.io/en/latest/messaging.html#copytoglobals.
+
+- Request (from UI to kernel)
+
+  ```json
+  {
+    'type': 'request',
+    'command': 'copyToGlobals',
+    'arguments': {
+      # the variable to copy from the frame corresponding to `srcFrameId`
+      'srcVariableName': str,
+      'srcFrameId': int,
+      # the copied variable name in the global scope
+      'dstVariableName': str
+    }
+  }
+  ```
+
+- Response (from kernel to UI)
+
+  ```json
+  {
+    'type': 'response',
+    'success': bool,
+    'command': 'setExpression',
+    'body': {
+      # string representation of the copied variable
+      'value': str,
+      # type of the copied variable
+      'type': str,
+      'variablesReference': int
+    }
+  }
--- a/debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
+++ b/debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
@@ -54,7 +54,7 @@ https://jupyter-client.readthedocs.io/en/latest/messaging.html#copytoglobals.
 
 - Response (from kernel to UI)
 
-  ```json
+  ```python
   {
     'type': 'response',
     'success': bool,
--- a/debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
+++ b/debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
@@ -1,0 +1,25 @@
+---
+title: Debugger support to copyToGlobals
+authors: Nicolas Brichet (@brichet)
+issue-number: xxx
+pr-number: xxx
+date-started: 2023-02-20
+---
+
+# Summary
+
+This JEP introduces a new field to the kernel debugger_info response. This new
+field will inform the UI that the debugger supports the `copyToGlobals` request.
+
+# Motivation
+
+The `copyToGlobals` request has been introduced in
+[ipykernel](https://github.com/ipython/ipykernel/pull/1055) and in
+[xeus-python](https://github.com/jupyter-xeus/xeus-python/pull/562) to copy a local
+variable to the global scope during a breakpoint. It would be useful to inform the
+UI if this is supported by the kernel before displaying the corresponding menu entry.
+
+# Proposed Enhancement
+
+We propose to add a new `copyToGlobals` boolean field to the `debugger_info`
+response which will inform the UI that this request is supported.
--- a/debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
+++ b/debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
@@ -38,7 +38,7 @@ https://jupyter-client.readthedocs.io/en/latest/messaging.html#copytoglobals.
 
 - Request (from UI to kernel)
 
-  ```json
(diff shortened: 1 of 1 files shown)


JohanMabille:
+1 on this, it has a very limited impact on the protocol and is backward compatible.

JohanMabille (approved):

debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
```diff
@@ -0,0 +1,25 @@
...
+
+The `copyToGlobals` request has been introduced in
+[ipykernel](https://github.com/ipython/ipykernel/pull/1055) and in
+[xeus-python](https://github.com/jupyter-xeus/xeus-python/pull/562) to copy a local
+variable to the global scope during a breakpoint. It would be useful to inform the
+UI if this is supported by the kernel before displaying the corresponding menu entry.
+
+# Proposed Enhancement
+
+We propose to add a new `copyToGlobals` boolean field to the `debugger_info`
```
krassowski, line 24: Would not it be more in line with DAP to advertise it via [`capabilities`](https://microsoft.github.io/debug-adapter-protocol/specification#Types_Capabilities) of [`initialize`](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Initialize) request response? For example, "copy to clipboard" can be implemented relying on `supportsClipboardContext?: boolean;`.

I see that this would require modifying the message in flight, but it seems cleaner on the API side.

krassowski: I guess the same goes for `richRendering` (which we have not codified so far).

brichet: Thanks @krassowski 

These "capabilities" are really kernel capabilities, and not debugger capabilities. I'm more in favor of separating the 2 types of capabilities, for clarity. An other solution could be to add a `capabilities` object in the `debugInfo` response, but that would not be backward compatible.

If I am not mistaken, all requests/responses defined in DAP are only forwarded by the kernel. It is just an interface between UI and debugger. Changes to the DAP message format will only affect the UI and not the kernels, making maintenance easier.

JohanMabille: > Would not it be more in line with DAP to advertise it via [`capabilities`](https://microsoft.github.io/debug-adapter-protocol/specification#Types_Capabilities) of [`initialize`](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Initialize) request response? For example, "copy to clipboard" can be implemented relying on `supportsClipboardContext?: boolean;`.

I'm not sure. The idea until now was to extend the existing DAP by adding new messages, not by modifying the existing messages or types used to compose the messages. Adding new capabilities to an existing message would mean that we need to introduce a new interface (inheriting form the Capabilities one) and replace the InitializeResponse spec with our own. From a protocol side, it is easier to maintain a clear separation between the DAP original specification, and our adds to it.

JohanMabille:
Additional context on this: the [copyToGlobals](https://jupyter-client.readthedocs.io/en/latest/messaging.html#copytoglobals) request was added to the Jupyter Debugger Protocol and implemented in [xeus-python](https://github.com/jupyter-xeus/xeus-python/pull/562) and [ipykernel](https://github.com/ipython/ipykernel/pull/1055). However, this request:
- is not part of the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol//specification.html)
- might be hard if not impossible to implement for some languages.

Therefore, a solution is to make it optional, like the [richInspectVariables](https://jupyter-client.readthedocs.io/en/latest/messaging.html#richinspectvariables) request, and avertise for its availabitily in the [debugInfo](https://jupyter-client.readthedocs.io/en/latest/messaging.html#debuginfo) request.

fcollonval:
Thanks for this JEP @brichet 

I would suggest describing in the JEP the request and response format of this new message; adding the section _Reference-level explanation_ seems the right place for that.

brichet:
> Thanks for this JEP @brichet
> 
> I would suggest describing in the JEP the request and response format of this new message; adding the section _Reference-level explanation_ seems the right place for that.

Thanks @fcollonval, I updated the description.

ivanov:
I am abstaining from this vote as I have no familiarity with the debugger. I have never had occasion to use it as a user, let alone dive into its implementation details as a developer.

brichet:
Some additional information to summarize this feature:

When the debugger stop on a breakpoint, some variables are in the global scope and some other can be in a local scope (in a function). Using the debugger tools only does not help for some introspection in complexe variables.

When the execution ends, the variable(s) in the local scope will not be reachable anymore.

We added a feature in some python kernels (ipykernel and xeus-python, see https://github.com/jupyter/enhancement-proposals/pull/93#issuecomment-1809107434) to be able to copy a variable from a local scope to a global scope during a breakpoint. This way, the variable(s) can be introspected after the code execution, since it still exists in the global scope.

The aim of this JEP is only to send a flag to Jupyterlab (or other UI), whether this feature is supported or not by the kernel.

debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
```diff
@@ -0,0 +1,69 @@
...
+
+If the feature is supported, the kernel must provide a function for copying a variable
+from a local scope to the global scope.
+The communication between the UI and the kernel (request - response) will have the
+structures described at
+https://jupyter-client.readthedocs.io/en/latest/messaging.html#copytoglobals.
+
+- Request (from UI to kernel)
+
+  ```json
```
gabalafou, line 41: The syntax highlighter is unhappy because JSON does not support single quotes nor comments

```suggestion
  ```python
```

debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
```diff
@@ -0,0 +1,69 @@
...
+      'srcFrameId': int,
+      # the copied variable name in the global scope
+      'dstVariableName': str
+    }
+  }
+  ```
+
+- Response (from kernel to UI)
+
+  ```json
```
gabalafou, line 57:
```suggestion
  ```python
```

debugger-info-copy-to-globals/debugger-info-copy-to-globals.md
brichet: Thanks @gabalafou

JohanMabille:
6 yes, 3 abstains, 2 no votes, let's merge!

PR #97, opened 2023-03-08
Add JEP for adding $schema to notebook format

filipsch:
This JEP proposes to add a new top-level field, `$schema` to the notebook JSON, as such updating the notebook JSON schema. This new field deprecates `nbformat` and `nbformat_minor`.

I skipped the step of creating a GitHub issue and deciding it's a JEP in this repository after discussing with @fcollonval. There was broad consensus about this change and the fact that it's a JEP during the notebook format workshop held in Paris (Feb 28 - Mar 2), and thought it okay to file a PR straight away. I will be the shepherd.

- @echarles 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @fcollonval 
  - [x] Yes
  - [ ] No 
  - [ ] Abstain
- @ibdafna 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @isabela-pf 
  - [ ] Yes 
  - [ ] No
  - [ ] Abstain
- @ivanov
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @JohanMabille 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @minrk
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @rpwagner 
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @SylvainCorlay
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @willingc
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @Zsailer 
  - [ ] Yes
  - [ ] No 
  - [ ] Abstain

### DIFF
--- a/97-add-schema/add-schema-to-notebook-format.md
+++ b/97-add-schema/add-schema-to-notebook-format.md
@@ -1,0 +1,117 @@
+---
+title: Add `$schema` to notebook format
+authors: Jason Grout (@jasongrout), Angus Hollans (@agoose77),
+Nicholas Bollweg (@bollwyvl), Filip Schouwenaars (@filipsch)
+issue-number: xxx
+pr-number: 97
+date-started: 2023-03-01
+---
+
+# Summary
+    
+We propose to add a new top-level field, `$schema` to the notebook JSON, as such updating the notebook JSON schema. This new field deprecates `nbformat` and `nbformat_minor`.
+
+# Motivation
+
+Today, `nbformat` and `nbformat_minor` specify the JSON schema a notebook should adhere to (for example [4.5](https://github.com/jupyter/nbformat/blob/main/nbformat/v4/nbformat.v4.5.schema.json)). Since this approach was adopted in the Jupyter ecosystem, the JSON schema standard has evolved and Jupyter's approach is no longer in line with it. Other than following standards being the right thing to do, bringing the notebook format back in line with the current JSON Schema spec opens it up to the rich tooling that exists around JSON schema validation today.
+
+# Guide-level explanation
+
+The introduction of a new `$schema` top-level property takes precedence over the existing `"nbformat"` and `"nbformat_minor"` properties that specify the notebook format.
+
+Example of notebook in 4.5 format:
+    
+```
+{
+    "nbformat": 4
+    "nbformat_minor": 5
+    "metadata": { ... }
+    "cells": [ ... ]
+}
+```
+    
+Example of notebook in 4.6 format:
+    
+```
+{
+    "$schema": "https://jupyter.org/schema/notebook/4.6/notebook-4.6.schema.json"
+    "nbformat": 4
+    "nbformat_minor": 6
+    "metadata": { ... }
+    "cells": [ ... ]
+}
+``` 
+
+It is REQUIRED that the top-level `$schema` be a canonical URI that refers to the Jupyter Notebook formats, i.e. `$schema` cannot be an arbitrary URI to a valid notebook schema, but instead must be useable as a token that uniquely identifies that schema, e.g.:
+
+Valid `$schema` URI:
+
+```json
+{
+    "$schema": "https://jupyter.org/schema/notebook/4.6/notebook-4.6.schema.json"
+    ...
+}
+```
+
+Invalid `$schema` URI:
+
+```json
+{
+    "$schema": "https://jupyter.org/schema/../schema/notebook/4.6/notebook-4.6.schema.json"
+    ...
+}
+```
+
+# Reference-level explanation
+
+JSON Schema changes:
+
+```
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Jupyter Notebook v4.5 JSON schema.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": ["$schema", "nbformat", "nbformat_minor", metadata", "cells"], // CHANGED
+  "properties": {
+     "$schema": {
+        "description": "JSON schema the notebook should adhere to",
+        "type": "string"
+      }
+    "metadata": { <no changes > }
+    "nbformat_minor": {
+      "description": "Notebook format (minor number). Incremented for backward compatible changes to the notebook format.",
+      "type": "integer",
+      "minimum": 6,
+      "deprecated": true   // CHANGED
+    },
+    "nbformat": {
+      "description": "Notebook format (major number). Incremented between backwards incompatible changes to the notebook format.",
+      "type": "integer",
+      "minimum": 4,
+      "maximum": 4
+      "deprecated": true   // CHANGED
+    },
+    "cells": [ <no changes> ]
+  },
+  "definitions": { <no changes> }
+}
+```
+
+# Rationale and alternatives
+
+Not doing this will leave the Jupyter notebook format in a non-standard state.
+
+# Prior art
+
+[JSON Schema](https://json-schema.org/) is a widely adopted declarative language that annotates and validates documents, so it's the obvious candidate to adhere to.
+
+# Unresolved questions
+
+- Code to upgrade from and downgrade to 4.5 still needs to be written.
+Some exploration has been done by Nick Bollweg in [this gist](https://gist.github.com/bollwyvl/a6e1ae13125f01ff04edf121e30a462a).
+- We want to deprecate `nbformat` and `nbformat_minor` in favor of `$schema` but we also need to ensure old clients can still work with notebooks in this new schema, so `nbformat` and `nbformat_minor` are still required. What's the path here? Major version update?
+
+# Future possibilities
+
+This work paves the way for [``Add `extraSchemas` to notebook format`` JEP](https://hackmd.io/9QZ8YibfQHm9l1B6JPSQsg?both), which will be submitted as a separate JEP soon.
--- a/97-add-schema/add-schema-to-notebook-format.md
+++ b/97-add-schema/add-schema-to-notebook-format.md
@@ -19,7 +19,7 @@ Today, `nbformat` and `nbformat_minor` specify the JSON schema a notebook should
 The new required `$schema` top-level property refers to a JSON Schema that validates the current notebook. During a deprecation period, `$schema` takes precedence over the existing `"nbformat"` and `"nbformat_minor"` properties that specify the notebook format. There will be a one-to-one mapping between `$schema` and the `(nbformat, nbformat_minor)` pair, which should follow a URI template, e.g.:
 ```json
 {
-  "$schema": "https://jupyter.org/schema/notebook/notebook-{nbformat}.{nbformat_minor}.schema.json",
+  "$schema": "https://schema.jupyter.org/notebook/v{nbformat}.{nbformat_minor}/notebook.json",
   "nbformat": 4,
   "nbformat_minor": 6
 }
@@ -41,7 +41,7 @@ The following changes are made to the existing v4.5 schema:
 -  "$schema": "http://json-schema.org/draft-04/schema#",
 -  "description": "Jupyter Notebook v4.5 JSON schema.",
 +  "$schema": "https://json-schema.org/draft/2020-12/schema",
-+  "$id": "https://jupyter.org/schema/notebook/4.6/notebook-4.6.schema.json",
++  "$id": "https://schema.jupyter.org/notebook/v{nbformat}.{nbformat_minor}/notebook.json",
 +  "description": "Jupyter Notebook v4.6 JSON schema.",
    "type": "object",
    "additionalProperties": false,
@@ -81,7 +81,7 @@ The following changes are made to the existing v4.5 schema:
 ```json
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://jupyter.org/schema/notebook/4.6/notebook-4.6.schema.json",
+  "$id": "https://schema.jupyter.org/notebook/v{nbformat}.{nbformat_minor}/notebook.json",
   "description": "Jupyter Notebook v4.6 JSON schema.",
   "type": "object",
   "additionalProperties": false,
@@ -563,7 +563,7 @@ Valid `$schema` URI:
 
 ```json
 {
-    "$schema": "https://jupyter.org/schema/notebook/4.6/notebook-4.6.schema.json"
+    "$schema": "https://schema.jupyter.org/notebook/v{nbformat}.{nbformat_minor}/notebook.json"
     ...
 }
 ```
@@ -572,7 +572,7 @@ Invalid `$schema` URI:
 
 ```json
 {
-    "$schema": "https://jupyter.org/schema/../schema/notebook/4.6/notebook-4.6.schema.json"
+    "$schema": "https://schema.jupyter.org/notebook/../notebook/v{nbformat}.{nbformat_minor}/notebook.json"
     ...
 }
 ```


willingc:
@MSeal @rgbkrk Please review. It looks fine to me if a migration path from old to new formats so notebooks from different nbformat versions can be executed.

97-add-schema/add-schema-to-notebook-format.md
```diff
@@ -0,0 +1,117 @@
...
+}
+```
+
+# Reference-level explanation
+
+JSON Schema changes:
+
+```
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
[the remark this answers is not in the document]
agoose77: We will also need to introduce support for this new keyword in the existing schemas. I don't know of a reason _not_ to bump to 2020 draft, besides any risk of existing tooling not working.

jasongrout:
Can we add what the proposed changes are to the schema that removes the deprecated attributes, to have an idea of what the resolution of the deprecated properties looks like?

westurner:
Please consider YAML-LD (JSON-LD) in naming the attribute `$schema` 

Given that Linked Data is ideal for science publishing and the internet, as explained and justified by https://5stardata.info/ 

Eventually, I and I believe also @bollwyvl TODO argue that nbformat should have a JSON-LD Context which would make `.ipynb` transformable to RDF; in order to botj
- publish Linked Data from nbformat notebooks
- publish nbformat notebooks as JSON-LD Linked Data 

Eventually, 
- https://github.com/jupyter/nbformat/issues/44
- https://github.com/lexiq-legal/pydantic_schemaorg validates Schema.org JSON-LD with jsonschema

That's out of scope for this issue, but FEIW the YAML-LD *Convenience Context* does map a bunch of things that start with `$` to their `@` equivalents in JSON-LD and `$schema` may or may not be confusing when working with nbformat as YAML-LD: 
```json
{
  "@context": {
    "$base": "@base",
    "$container": "@container",
    "$direction": "@direction",
    "$graph": "@graph",
    "$id": "@id",
    "$import": "@import",
    "$included": "@included",
    "$index": "@index",
    "$json": "@json",
    "$language": "@language",
    "$list": "@list",
    "$nest": "@nest",
    "$none": "@none",
    "$prefix": "@prefix",
    "$propagate": "@propagate",
    "$protected": "@protected",
    "$reverse": "@reverse",
    "$set": "@set",
    "$type": "@type",
    "$value": "@value",
    "$version": "@version",
    "$vocab": "@vocab"
  }
}
```

- https://github.com/RDFLib/rdflib-jsonld/issues/110
  - https://json-ld.github.io/yaml-ld/spec/#example-convenience-context

`$schema` is not on the list.

agoose77:
@westurner the `$schema` top-level property is already prior-art for declaring that a document conforms to a JSON Schema. If we use a different property here, we lose the ability to have a large class of validators understand our document. We've touched on RDF/JSON-LD in our weekly meetings, which you're encouraged to join!

tonyfast:
currently, the top level notebook schema does not allow for any additionalProperties defined in the container, so we can't have any LD  `@context`. we're hoping to introduce `@context` as a top level key in future schema. there are likely a few proposals between this JEP and an `@context` proposal. so this is on folks minds, but we decided to defer linked data proposals until some prior JEPs are accepted. advancing the schema will be mean good things for our ability to write linked data contexts.

tonyfast:
@jupyter/software-steering-council we are working on a draft to present to y'all for the JEP. yesterday we were wondering what to expect with the process. is there any way someone can outline what the process will look like so we can plan our work accordingly and set some deadlines?

rgbkrk:
This is so much more sensible than the incrementing numbers and awkward compatibility between notebook formats. I'm wholly on board. Thank you all so much for pushing forward with this approach.

rgbkrk (approved):

97-add-schema/add-schema-to-notebook-format.md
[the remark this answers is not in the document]
fcollonval: I would say for the newer version it is acceptable to bump the schema to the latest draft after ensuring there is no backward incompatibility between the current version based on draft 04 and the new draft - as far as I know some breaking changes are induced when upgrading from [draft 04 to draft 06](https://json-schema.org/draft-06/json-schema-release-notes.html#backwards-incompatible-changes).

> From a quick look it seems ok

Additionally if we are to bump the JSON Schema draft, I would switch all `enum` with single value to the new `const` as used for the nbformat version numbers (see mainly `cell_type` and `output_type`).

fcollonval:
Thanks all for the great discussion.

>  is there any way someone can outline what the process will look like so we can plan our work accordingly and set some deadlines?

For my reading there are three opened questions:
- [ ] What should be the JSON Schema Draft version?
   - [ ] Should we annotate the `nbformat_minor` and `nbformat` as deprecated (if we use draft 2019 or later)?
- [ ] Should we switch more `enum` with single value to `const` as the newer draft allows that (this will ease the understanding).

And I'm unclear about the following comment of @agoose77 :

> We will also need to introduce support for this new keyword in the existing schemas

Which new keyword are we speaking about?

---

To get validation (from the SSC), the easiest would be to resolve all pending questions and then ping the SSC that this is ready for approval. If some questions are left opened, I would recommend summarizing them in a comment with the possible solutions. Then ping the SSC that will have to figure out how to move forward.

agoose77:
> What should be the JSON Schema Draft version? 
At _least_ 2019-09. I'd be curious to know whether there are downsides to just jumping straight to 2020-12.

> Should we annotate the nbformat_minor and nbformat as deprecated
Yes, I think so.

>  Should we switch more enum with single value to const as the newer draft allows that (this will ease the understanding).

Yes, I think so.

> Which new keyword are we speaking about?

Actually, this is something I wanted to follow up with @jasongrout on. Due to the fact that our existing schemas have `additionalProperties: false`, no document with the `$schema` top-level property will be considered valid. Right now, this doesn't cause a hard-failure with `nbformat`; the validator complains about a validation error, but ultimately loads notebooks with additional properties. My understanding of our deprecation process is whilst we _will_ update `nbformat` so that it always uses `$schema` if it finds it, we should keep the `nbformat` properties in these transition notebooks so that out-of-date `nbformat` libraries / other validators have a chance at being able to read the notebook. 
i.e., if they're not strictly rejecting the document outright due to the new `$schema` property, then they will have sufficient information to know that it's nbformat 4.

97-add-schema/add-schema-to-notebook-format.md
```diff
@@ -0,0 +1,599 @@
...
+{
+  "$schema": "https://jupyter.org/schema/notebook/notebook-{nbformat}.{nbformat_minor}.schema.json",
+  "nbformat": 4,
+  "nbformat_minor": 6
+}
+
+```
+After the deprecation period expires, a future JEP will remove these `nbformat` and `nbformat_minor` properties from the notebook schema. These properties are retained to permit legacy notebook consumers to read notebooks authored during this deprecation period.
+
+The addition of the `$schema` property removes a level of indirection between the notebook and the schema against which it is invalidated. It also guarantees that the schema against which it is validated is invariant with respect to time; the schema URI should refer to an immutable document.
```
rpwagner, line 30: > `the schema URI should refer to an immutable document.`

It would be good to define where the canonical version of the schema document is stored. This may already be defined somewhere else, in which case that information could be referenced here.

While it's not necessary, the could be hosted at the URL represented by the URI. That is, `https://jupyter.org/schema/notebook/notebook-{nbformat}.{nbformat_minor}.schema.json` would resolve to the actual schema document. Or, it could be tied a GitHub repo, branch, and tag.

agoose77: Agreed, this is something we've touched upon in the meetings. It's my feeling that we haven't wanted to define that in this JEP (to avoid taking on too many responsibilities). As of right now, the schemas used to validate notebooks are stored in the nbformat repository / wheels. I don't *think* they're hosted on a standalone URL, but I've not checked!

rpwagner: I wouldn't call this a strong need or requirement. But when I see a schema declaration, my instinct is to ask where I can find the actual document and the statement regarding an immutable document reinforces that. If this isn't the right time to define the process for hosting the schema, then I would suggest just documenting the current location as information for the reader.

bollwyvl: >  could be tied a GitHub repo, branch, and tag.

`jupyter.org` will, in all likelihood, remain under control of Project Jupyter, while GitHub could pull a Docker, Inc., and make things much more challenging.

Further, and not explicitly stated (again to avoid any scope creep): it must _not_ be an expectation that a validating tool will need to (or even be able to) fetch the schema in order to validate it.

While it's my strong feeling the whole family of current and future Jupyter schema should all have an "official", inspectable URL together, with unified tooling for generating human-readable documentation of the schema, generated, lightweight header/typing packages should obviate much of the need for "go grab something off the internet at runtime".

Today, `nbformat` publishes canonical packages on `pypi.org` and `npmjs.com`, which is a great start! But really every language community that _wants_ representation in Jupyter should be able to propose and maintain lightweight packages, and get up-to-date packages.

Here's a quick [strawman sketch](https://discourse.jupyter.org/t/sustainability-of-the-ipynb-nbformat-document-format/3051/5) of what something like that might look like.

westurner:
> * Should we annotate the `nbformat_minor` and `nbformat` as deprecated (if we use draft 2019 or later)?

When [W3C SHACL] validation for Linked Data notebooks becomes the norm (because Linked Data Notebook outputs are most practically validated as Linked Data with Shapes and Constraints), then the (Uproperty for the version of the SHACL validation document would need to supersede `$schema` y

westurner:
> currently, the top level notebook schema does not allow for any `additionalProperties` defined in the container, 

Is JSONschema with `additionalProperties: false` fundamentally incompatible with JSON-LD?

> so we can't have any LD `@context`. we're hoping to introduce `@context` as a top level key in future schema. 

When the versioned URI and contents of the `@context` attribute change, does the nbformat major or minor version need to change?

> there are likely a few proposals between this JEP and an `@context` proposal. so this is on folks minds, but we decided to defer linked data proposals until some prior JEPs are accepted. advancing the schema will be mean good things for our ability to write linked data contexts.

nbformat is older than jsonschema, and may outlast jsonschema draft n, so a separate version string that doesn't change between implementations would be great

tonyfast:
> Is JSONschema with `additionalProperties: false` fundamentally incompatible with JSON-LD?

notebook documents, the serialized version of someone's notebook, is fundamentally incompatible with JSON-LD. we can add `@context` and `@graph` or any other json-ld key into the `metadata` properties because they have permissive keys. the top level notebook document is much more strict. in fact, without this JEP, `$schema` is not something that can exist in a serialized notebook document because `additionalProperties` is false.

> When the versioned URI and contents of the `@context` attribute change, does the nbformat major or minor version need to change?

this is a good consideration, as we deprecate `nbformat` and `nbformat_minor` we'll have to increment with each version until the deprecate. we have ongoing discussions about how to handle mismatched `$schema` and `nbformat` keys, likely `$schema` takes precedence.

> nbformat is older than jsonschema, and may outlast jsonschema draft n, so a separate version string that doesn't change between implementations would be great for backward compatibility

wow, you're right! `nbformat` does predate `jsonschema`, it seems `v3` is the first version to rely on draft04. that was a fun dig into history. anyway, current json schema efforts seem to be a well supported community and they are rigorous in their changing their versions. `nbformat` will update the json schema draft it is based on less than we will update our own schema versions. 

we've been spending a lot of time discussing backwards compatibility, and how to handle that best. on going work...



> When [W3C SHACL] validation for Linked Data notebooks becomes the norm (because Linked Data Notebook outputs are most practically validated as Linked Data with Shapes and Constraints), then the (URI-namespaced) property for the version of the SHACL validation document would need to supersede

a shacl context for notebook schema will undoubtedly show up in the future. the `nbformat` schemas serve as valuable interfaces defining linked data contexts. for example, `nbformat` could be mapped to `shacl` using a context like:

```json
 {"@vocab": "https://github.com/jupyter/nbformat/blob/main/nbformat/v4/nbformat.v4.5.schema.json#", "@base": "http://www.w3.org/ns/shacl#](http://www.w3.org/ns/shacl#"}
```

tonyfast:
attaching notes from last week's meeting. see folks tomorrow.

## April 4rd, 2023

| Name | Affiliation | GitHub  |
| ------------------ | ----------- | ------------- |
| tonyfast | | @tonyfast |
| jeremy ravenal  | naas | @jravenel |
| Angus Hollands   | Princeton University| @agoose77 |
| Afshin T. Darian | QuantStack | @afshin | 

### Agenda

* schema provide solutions for validation and ui
* angus on extra schemas
    * platform to design a different notebooks that will allow different input cells.
    * extension authors can encode some other validation logic
    * mainly useful for the front end.
* what is the history of this meeting?
    * spun out of a workshop that discussed modifying the notebook format
    * we talked about different ways to create new cells types: is there one code cell or many different cells?
    * there is no nice way for myst to store metadata, there is no way to enshrine metadata in the schema
* naas - push notebooks to production seamlessly. package software, data, and chats.
* sell things with demos
  - raw cell bolt-on
  - add a MIME / text entry widget to the cell view
    - specify the mimetype of the contents
  - add a custom renderer based upon mime type
    - HTML
    - SVG
    - Form generation (one way, code generation, hacky!)
  - Execute cell has two steps
    - Two steps, one "executes", another "renders"
  - ability to select output mimetype as well?
* we talked a lot about cell types
    * use raw cells for another cell type
    * use the kernelspec in cells to identify cell actions
* steps to getting JEP accepted
    * what implementation will need updating? (open question)
* Split these conversations:
  * `$schema` - uncontentious
  * `extraSchemas` - motivates extension validation, needs discussion
  * `@context`, annotation, etc. additional discussions!

97-add-schema/add-schema-to-notebook-format.md
Zsailer: FYI: https://github.com/jupyter/enhancement-proposals/issues/107

tonyfast:
## April 25th, 2023

| Name | Affiliation | GitHub  |
| ------------------ | ----------- | ------------- |
| tonyfast | | @tonyfast |
| Nick Bollweg | GTech | @bollwyvl |


### Agenda

*  we collaborated on another JEP for [jupyter.org subdomain and repository for publishing schemas](https://hackmd.io/ujzmp3tSQJqP24H6Jm2s2w). it was submitted by @zsailor https://github.com/jupyter/enhancement-proposals/pull/108

tonyfast:
## May 2nd, 2023

| Name | Affiliation | GitHub  |
| ------------------ | ----------- | ------------- |
| tonyfast | | @tonyfast |
| Angus Hollands   | Princeton University| @agoose77 |


### Agenda

* recently, some JEPs have been triaged with __Under Discussion (RFC)__. the following outstanding JEPs should be marked the same. 
    * https://github.com/jupyter/enhancement-proposals/pull/97
    * https://github.com/jupyter/enhancement-proposals/pull/101

> is there someone with the proper rights to label these JEPs?

JohanMabille (approved):

SylvainCorlay (approved):

fcollonval:
@/all (but especially @jupyter/software-steering-council) in 0871ad1 I updated the schema URI to align with JEP #108; i.e. from `https://jupyter.org/schema/notebook/notebook-{nbformat}.{nbformat_minor}.schema.json` to `https://schema.jupyter.org/notebook/v{nbformat}.{nbformat_minor}/notebook.json`

For this particular URI I did not use a subproject (as [allowed](https://jupyter.org/enhancement-proposals/108-jupyter-subdomain-for-schemas/jupyter-subdomain-for-schemas.html) by the JEP). Let me know if it needs further changes.

ivanov (approved):

fcollonval:
The vote is now closed with the results:

**In favor**: 8
**Against**: 0
**Abstention**: 0
**No vote**: 3

--> In light of those results, this JEP is accepted.

Issue #98, opened 2023-03-10
Pre-proposal: Specify the Markdown cell's markdown flavor


fcollonval:
> This proposal was developed during Jupyter Community Workshop on the Notebook format (Paris, 28 Feb - 2 Mar).

This is a proposition to add a new key to markdown cells defining precisely the flavor of markdown used in the cell source. The associated rendered content should be stored for compatibility with the tooling ecosystem. As a side effect, it forces us to clarify the current default markdown flavor that should be used as fallback when a client does not support the specified flavor.

The focus of the associated JEP is primarily on describing current behaviour.

During the workshop we wrote a draft JEP to ease structuring the goals and discussion. You can find it on the following draft PR:

fcollonval:
The opened questions already seen are:

- Should the default be `text/markdown`? And the fallback be Original Markdown?
- Should clients update the mimetype key to match what they are rendering/supporting after editing?
- Should frontend only change a markdown `mimetype`  in an edited cell, and only change that cell?
- Is it ok for a notebook to have cells in a different format?
- Should it be specified as document level key and interpreted as a **hint**, i.e. a frontend serializing the notebook could write a mimetype into the notebook to communicate how clients **should try** to interpret markdown cells and then make best efforts to do so (similar to the kernel information).
- Should there be a single output or an array of outputs to align better with code cells? In the second case:
  - what to do if there are more than 1 output
  - should we force the output type to a single value? And if yes, the most appropriate seems [display_data](https://github.com/jupyter/nbformat/blob/16b53251aabf472ad9406ddb1f78b0421c014eeb/nbformat/v4/nbformat.v4.5.schema.json#L331)

bollwyvl:
Some thoughts:

> Should the default be text/markdown?

For users' sake, the default should be as backwards-compatible as possible, therefore probably `text/dollar-math+x-gfm` or something. 

>  And the fallback be Original Markdown?

The original author of Markdown has granted that fenced code blocks and tables pretty much were grossly missing, and it is widely known the reference perl implementation contains bugs and onconsistences that can [never be fixed](https://www.hyrumslaw.com/).

The only _real_ fallback is `text/plain` (and ideally _in what human language_), though some of the bespoke markdown formats are starting to get _really_ far from either plain text _or_ a human language.

> Should clients update the mimetype key to match what they are rendering/supporting after editing?

[...]

Issue #95, opened 2023-03-06
Rethinking notebook cell types


fcollonval:
> This topic is one of the output from the Jupyter Community Workshop on the notebook file format.

This is an issue to synchronize and invite any interested body to take part to the discussion of the major new file format structure.

A weekly call has been set up for Tuesday 4:00 PM GMT - https://meet.google.com/sac-mgxn-ndg

The notes from the workshop are available at https://docs.google.com/document/d/1DMMUOYEhFxoAEKITOrCUK9x0vkTy68mfZ9clof3UrMc

The meeting notes are available: Coming as soon as we start the meeting :wink:

willingc:
Thanks for posting @fcollonval. FYI @MSeal @rgbkrk.

agoose77:
I won't be able to make this week's meeting! I hope it proves fruitful :)

krassowski:
I just wanted to highlight that any changes here would ideally consider what should happen to trust. Currently nbformat only allows for transient `trusted` metadata to be declared on code cells which means that:
- markdown cells are never trusted and always sanitised (which is part of current Notebook security model)
- new interfaces like JupyterLab for a long time had trouble producing a JSON dump satisfying nbformat, which lead to trust being wrongly lost as signatures were lost upon reformatting by nbformat (that was done in the background)

Two previously proposed ideas on improving trust/signature mechanism are:
- split current "catch all" signature into multiple certificates (https://github.com/jupyter/nbformat/issues/296#issuecomment-1232104286):
   - one for integrity of the document ("this notebook was not modified")
   - another mechanism for trusting outputs of cells (or maybe cells in general)
- implement granular per-cell, per-MIME type trust metadata (https://github.com/jupyterlab/jupyterlab/issues/12889#issuecomment-1435993926) for example accommodating for users who may want to trust all CSS and HTML but not JS (especially relevant for more powerful markdown cells that would blend some of the distinctions); this could look like `trust: ['text/markdown', 'text/html', 'text/my-custom-language']`

echarles:
> markdown cells are never trusted and always sanitised (which is part of current Notebook security model)

[jupyterlab-myst](https://github.com/executablebooks/jupyterlab-myst) has now the ability to evaluate python/kernel code inserted into markdown cells, and overrides the default notary handler to deal with that. The evolution should imho add trust notion to any cell (code, markdown or any type that would come)

PR #99, opened 2023-03-10
First draft markdown flavor JEP

fcollonval:
This draft PR is to share the output of the Notebook format workshop on specifying the Markdown flavor description in notebook file.

Please discussed this proposal in the associate issue: https://github.com/jupyter/enhancement-proposals/issues/98.

### DIFF
--- a/##-markdown-flavor/markdown-flavor.md
+++ b/##-markdown-flavor/markdown-flavor.md
@@ -1,0 +1,257 @@
+---
+title: Specify the Markdown cell’s markdown flavor
+authors: Frederic Collonval, Steve Purves and Vincent Vankrunkelsven
+issue-number: 98
+pr-number: 99
+date-started: 2023-03-10
+---
+
+# Summary
+
+This is a proposition to add a new key to markdown cells defining precisely the flavor of markdown used in the cell source. The associated rendered content should be stored for compatibility with the tooling ecosystem. As a side effect, it forces us to clarify the current default markdown flavor that should be used as fallback when a client does not support the specified flavor.
+
+The focus of this JEP is primarily on describing current behaviour.
+
+# Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected outcome? -->
+
+The flavor of markdown cells in the notebook are currently not defined, it is not possible to describe precisely how markdown will be rendered or which syntax is supported.
+
+Currently it might be mostly GFM.
+Better define the MD flavor to set user expectations properly
+Enable rendering of other markdown flavors by extension
+Understand what the fallback would be
+
+For reference, the current Markdown flavor is described in the documentation as:
+
+> Markdown cells are used for body-text, and contain markdown, as defined in [GitHub-flavored markdown](https://help.github.com/articles/github-flavored-markdown), and implemented in [marked](https://github.com/chjj/marked).  
+https://nbformat.readthedocs.io/en/latest/format_description.html#markdown-cells
+
+> The Notebook webapp supports GitHub flavored markdown.  
+https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Working%20With%20Markdown%20Cells.html#GitHub-flavored-markdown
+
+Those descriptions are inaccurate especially since GitHub is quite active to extend the rendered Markdown; see for example [Creating diagrams](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams).
+
+# Guide-level explanation
+
+<!-- Explain the proposal as if it was already implemented and you were
+explaining it to another community member. That generally means:
+
+- Introducing new named concepts.
+- Adding examples for how this proposal affects people's experience.
+- Explaining how others should *think* about the feature, and how it should impact the experience using Jupyter tools. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing Jupyter members and new Jupyter members.
+
+For implementation-oriented JEPs, this section should focus on how other Jupyter
+developers should think about the change, and give examples of its concrete impact. For policy JEPs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms. -->
+
+This proposal will introduce a new key `mimetype` to the markdown cell. It is not mandatory to allow backward compatibility. And therefore, this proposal is introducing a _fallback value_ (TBD).
+
+Then the markdown cell can have a single `output` storing the rendered content as a mime bundle. If for example the mime bundle is providing text/html rendered content, a tool like nbconvert could inject that html directly when converting the notebook to HTML.
+
+Due to the markdown fragmentation, a side effect of this JEP is the need to define a fallback Markdown flavor. From what is out there, the best path seen is to create and maintain an integration test suite for the fallback supported syntax. This will clarify the supported syntax and the associated rendered HTML. This is similar to what [CommonMark](https://spec.commonmark.org/) and [GFM](https://github.github.com/gfm/) are defining.
+A starting point would be that [pull request](https://github.com/jupyterlab/benchmarks/pull/97). But moving it into [nbformat](https://github.com/jupyter/nbformat) repository. The open-source tools will need to be aligned - in particular the question of aligning Python based renderers like nbconvert and web based renderers like JupyterLab/Notebook must be tackled?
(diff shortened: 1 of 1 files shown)


##-markdown-flavor/markdown-flavor.md
```diff
@@ -0,0 +1,257 @@
...
+- Introducing new named concepts.
+- Adding examples for how this proposal affects people's experience.
+- Explaining how others should *think* about the feature, and how it should impact the experience using Jupyter tools. It should explain the impact as concretely as possible.
+- If applicable, provide sample error messages, deprecation warnings, or migration guidance.
+- If applicable, describe the differences between teaching this to existing Jupyter members and new Jupyter members.
+
+For implementation-oriented JEPs, this section should focus on how other Jupyter
+developers should think about the change, and give examples of its concrete impact. For policy JEPs, this section should provide an example-driven introduction to the policy, and explain its impact in concrete terms. -->
+
+This proposal will introduce a new key `mimetype` to the markdown cell. It is not mandatory to allow backward compatibility. And therefore, this proposal is introducing a _fallback value_ (TBD).
```
rgbkrk, line 50: This may allow going a bit more broad than I'd recommend, since mimetype can be a lot of unrelated and yet well specified types.

Conceptually though, calling it `markdown_format` would make sense. Limit it to an enum of possible variants:

* `marked`
* `commonmark`
* `gfm`
* `commonmark+latex`

##-markdown-flavor/markdown-flavor.md
```diff
@@ -0,0 +1,257 @@
...
+
+Then the markdown cell can have a single `output` storing the rendered content as a mime bundle. If for example the mime bundle is providing text/html rendered content, a tool like nbconvert could inject that html directly when converting the notebook to HTML.
+
+Due to the markdown fragmentation, a side effect of this JEP is the need to define a fallback Markdown flavor. From what is out there, the best path seen is to create and maintain an integration test suite for the fallback supported syntax. This will clarify the supported syntax and the associated rendered HTML. This is similar to what [CommonMark](https://spec.commonmark.org/) and [GFM](https://github.github.com/gfm/) are defining.
+A starting point would be that [pull request](https://github.com/jupyterlab/benchmarks/pull/97). But moving it into [nbformat](https://github.com/jupyter/nbformat) repository. The open-source tools will need to be aligned - in particular the question of aligning Python based renderers like nbconvert and web based renderers like JupyterLab/Notebook must be tackled?
```
rgbkrk, line 55: Rather than do it per cell, I'd aim for the notebook document _itself_ to have the markdown format specified at the top level. There are `text/markdown` outputs after all.

##-markdown-flavor/markdown-flavor.md
```diff
@@ -0,0 +1,257 @@
...
++             "properties": {
++                 "data": { "$ref": "#/definitions/misc/mimebundle" },
++                 "metadata": { "$ref": "#/definitions/misc/output_metadata" }
++             }
++         }
+      }
+  },
+```
+
+The structure of the mime type must follow the standard defined in [RFC7763](https://www.rfc-editor.org/rfc/rfc7763): `text/markdown;variant=<variant name>` (variant parameter is optional).
```
rgbkrk, line 125: TIL there's a way to declare a variant of markdown.

jjallaire: Registered variants are here: https://www.iana.org/assignments/markdown-variants/markdown-variants.xhtml

##-markdown-flavor/markdown-flavor.md
```diff
@@ -0,0 +1,257 @@
...
+    omitted, there is no hint; the interpretation is entirely up to
+    the receiver and context. This identifier is plain US-ASCII and
+    case-insensitive. To promote interoperability, identifiers can be
+    registered in the registry defined in Section 6. If a receiver
+    does not recognize the variant identifier, the receiver MAY
+    present the identifier to a user to inform him or her of it._
+
+The resulting fully specified `mimetype` would be `text/markdown;variant=GFM`
+
+The [RFC7763](https://www.rfc-editor.org/rfc/rfc7763.html) specifies a [registry provided by IANA](https://www.iana.org/assignments/markdown-variants/markdown-variants.xhtml).
```
rowanc1, line 204: Both `quarto` and `myst` have been added to the registry in the last few weeks based on follow ups from the notebook format meeting. If would be great to add them to the JEP!

##-markdown-flavor/markdown-flavor.md
```diff
@@ -0,0 +1,257 @@
...
+- Why is this choice the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+-->
+
+Specifying the markdown format should leverage the concept already used in the notebook document. It feels therefore natural to choose a mimetype to specify the source format. The notion of transient output exists already for code cell. Leveraging a similar structure for markdown cell sounds like the natural choice.
+
+Adding this feature to the notebook file has the risk of increasing incompatibility between clients. But this has already started as client extensions extending or replacing the markdown renderer in JupyterLab already exists:
+- [jupyterlab-myst](https://github.com/executablebooks/jupyterlab-myst): MyST functionality in Jupyter Lab
+- [jupyterlab-markup](https://github.com/agoose77/jupyterlab-markup): JupyterLab extension to enable markdown-it rendering, with support for markdown-it plugins
```
stevejpurves, line 149: This list has now expanded :D : https://www.npmjs.com/package/@quarto/jupyterlab-quarto

##-markdown-flavor/markdown-flavor.md
stevejpurves: Apat from being well established way to identify the content the interesting thing using `mimetype` here is thst. is could potentially be used to specify other content like YAML, or HTML or SQL which could be rendered appropriately in a "markdown" cell - maybe we should think of markdown cell more as an "input" cell.

Immediately opens up the case for notebooks with hetrogenous inputs (for better or worse) including markdown variants, or things like YAML, or SQL and for that content to be properly identified - I believe this content is all already in use in notebooks but not well supported

##-markdown-flavor/markdown-flavor.md
stevejpurves: also note the discussion on [this issue](https://github.com/jupyter/enhancement-proposals/issues/98)

krassowski:
I think that this proposal should ideally address the question of security model for proposed outputs. The current Jupyter Notebook  [security model](https://jupyter-notebook.readthedocs.io/en/stable/security.html#our-security-model) prohibits execution of JS and forces sanitation of HTML/CSS in all Markdown cells. Since `outputs` in this proposal are modelled on code cells (which can be trusted or not), would it be worth re-visiting the trust question?

Please see more thoughts on potential trust evolution in https://github.com/jupyter/enhancement-proposals/issues/95#issuecomment-1501176251.

fcollonval:
We have started looking at this at the SSC meetings. We have decided to give at least another 2 weeks of discussion before moving forward.

##-markdown-flavor/markdown-flavor.md
fcollonval: Thanks for the follow up @rowanc1 ? Would you mind providing guidance how the process to add such variant?

jjallaire: You can register a new variant by filling out this form: https://www.iana.org/form/protocol-assignment

jjallaire:
What is the current thinking around the precise content type of the fallback ("Rendered Markdown source")? This might be clear in the revised text of the proposal but I wanted to be sure. The scenarios I'm thinking offer for renderers like nbconvert or for notebook editors that don't know about the specified `mimetype`. In these cases it seems like editing tools should provide a standard Jupyter Markdown version of the content -- e.g. if a callout supported by MyST or Quarto is in the markdown, the fallback should be something suitable e.g. a markdown blockquote). The alternative would be to provide e.g. `text/html` however if you are attempting to render a PDF or DOCX this won't be helpful (whereas Jupyter Markdown will be).

I also wonder if given that notebook editors won't immediately be aware of this extension whether the actual `source` should be always written in Jupyter Markdown, and the source that uses an extended markdown variant be provide separately. This would allow "variant aware" editors to deal with the content properly but allow the rest of the ecosystem to keep working without changes?  This might already be anticipated in an aspect of the proposal I'm not appreciating, but in the case that its not is definitely worth considering.

PR #104, opened 2023-04-01
Reviewed JEP process

JohanMabille:
This is an update of the JEP process. Main changes are:

- the Shepherd is now optional
- the voting phase has been explicitly added
- An intermediate status has been added to help JEP triage and review
- the JEP process diagram has been updated according to these changes.

After this is reviewed and merged, we should make clear that JEP 29 is deprecated or outdated to avoid confusion (since there will be two JEPs describing the process). Another solution would be to open a PR to the JEP 29 that has already been merged instead of this new JEP, but I don't like this solution as it "erases" the history.

There is also an old description of the JEP process [here](https://github.com/jupyter/enhancement-proposals/blob/master/jupyter-enhancement-proposal-guidelines/jupyter-enhancement-proposal-guidelines.md) that we should replace with the last version of the JEP process (i.e. the doument from this PR when it is approved).

Looking forward to reading your feedbacks!

---

- @echarles 
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @fcollonval 
  - [x] Yes
  - [ ] No 
  - [ ] Abstain
- @ibdafna 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @isabela-pf 
  - [x] Yes 
  - [ ] No
  - [ ] Abstain
- @ivanov
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @JohanMabille 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @minrk
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @rpwagner 
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @SylvainCorlay
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @Zsailer 
  - [x] Yes
  - [ ] No 
  - [ ] Abstain


minrk (approved):

jep-process-v2/jep-process-v2.md
```diff
@@ -0,0 +1,164 @@
...
+
+1. A github issue on the Jupyter Enhancement Proposals repo is created that
+2. 1. Briefly outlines the proposal
+
+   2. Suggests reviewiers (optional) in addition to the SSC members.
+
+   3. Why it should be a JEP
+
+   4. 1. See the “JEP / Not-a-JEP Rubric” below.
+3. A *Shepherd* is identified to see the process through. Shepherds are assigned from a set of people actively involved in the project (Sotfware subprojects, SSC or EC, Standing committees and Working groups), based on the scope of the JEP. The Shepherd can be the author of the JEP.
```
fcollonval, line 78:
```suggestion
assigned from a set of people actively involved in the project (Sotfware subprojects, SSC or EC, Standing committees and Working groups)
```

As this restricts the people eligible as *Shepherd* should we have a fallback like the SSC will pick a shepherd or clarify how the shepherd is chosen. My concern is that if a JEP author is not well known member of the community, he/she may not understand what to do to find a shepherd.

JohanMabille: I agree, I'll add this fallback.

ivanov:
I pushed up some minor fixes, and on-board with making these changes, but how do we feel about adding a "Superseded" or "Replaced" status for JEPs? It seems like this is exactly the sort of thing we would want to apply to JEP29 as part of this.

Another alternative would be to have an "Active" status, which is [what Python does with PEPs that can change and be updated](https://peps.python.org/pep-0001/), like PEP 1.

jep-process-v2/jep-process-v2.md
```diff
@@ -0,0 +1,165 @@
+---
+title: Jupyter Enhancement Proposal
+authors: |
+  Jason Grout ([jason@jasongrout.org](mailto:jason@jasongrout.org)), Safia Abdalla ([safia@safia.rocks](mailto:safia@safia.rocks)), John Lam ([jflam@microsoft.com](mailto:jflam@microsoft.com)), Kevin M. McCormick ([mckev@amazon.com](mailto:mckev@amazon.com)), Pierre Brunelle ([brunep@amazon.com](mailto:brunep@amazon.com)), Paul Ivanov ([pi@berkeley.edu](mailto:pi@berkeley.edu))
+issue-number: 27
+pr-number: 29
+date-started: "2019-02-23"
```
ivanov, line 7: maybe add a "Date updated" here for this revision of the changes.

minrk:
> how do we feel about adding a "Superseded" or "Replaced" status for JEPs?

I think _both_ 'superseded' and 'active' states make sense, thanks for bringing those up!

If we have 'active' for process JEPs like 29, we may not need 'superseded' specifically as a state, and can do with adding a _note_ that folks looking at an accepted JEP may want to look at a later one. A formal 'superseded' state may not really solve a problem.

JohanMabille:
I finally took the time to get back to this; I've taken into account the last comments and reworked the JEP workflow. If everyone is happy with it, we could call a vote by the end of the week

29-jep-process/jep-process.md
```diff
@@ -87,15 +88,15 @@ The Shepherd decides if the JEP criteria have been met.
 
 ### Phase 2: RFC for the JEP
 
-Submission: The author submits an initial draft of the JEP as a PR to the JEP repository starting from the relevant template decided in the pre-proposal stage. The Shepherd assigns a number (the GitHub PR number? A sequential number? Perhaps the number isn't assigned until the JEP is merged?). The Shepherd assigns a Review Team.
+Submission: The author submits an initial draft of the JEP as a PR to the JEP repository starting from the relevant template decided in the pre-proposal stage. The Shepherd assigns a number (usually the Github PR number). The Shepherd assigns a Review Team.
```
davidbrochart, line 91:
```suggestion
Submission: The author submits an initial draft of the JEP as a PR to the JEP repository starting from the relevant template decided in the pre-proposal stage. The Shepherd assigns a number (usually the GitHub PR number). The Shepherd assigns a Review Team.
```

jep-process-v2/jep-process-v2.md
```diff
@@ -0,0 +1,170 @@
...
+
+- **There is one JEP repository, all Jupyter-governed projects must use it.**  To facilitate the easiest possible adoption and high visibility of ideas, a single JEP repository will be used.  Even if a JEP only applies to a single organization.
+
+- The JEP process **has multiple valid use cases**.  Each use case might have a slightly different expected workflow or base JEP template.  Some expected use cases include:
+
+  - Non-trivial feature proposals within a single component that would benefit from process.  (e.g., a non-trivial change to JupyterLab that would benefit from formal process within the JupyterLab project)
+  - Non-trivial features or improvements that span multiple projects.
+  - Any proposed changes to published APIs or core specifications (e.g., nbformat)
+  - Changes to the JEP process itself.
+  - Creating a new GitHub repo in one of the official Jupyter orgs
```
rpwagner, line 60: The example of creating a single *repo* within an existing GitHub org feels too specific for a JEP. However, the creation of a new GitHub *org* does need a more well-defined process to ensure that it has the appropriate ownership and configuration (e.g., 2FA). I suggest changing this line to simply `Creating a new official GitHub organization`.

Thinking further, it may be worth a JEP to clarify the creation and management of GitHub entities (orgs, repos, teams, etc.). I think the SSC is the place to handle that JEP, with additional review by the Governance Council.

jep-process-v2/jep-process-v2.md
```diff
@@ -0,0 +1,170 @@
...
+**Principles to follow**
+
+Below are a few example guidelines to follow when deciding if an idea should include
+a JEP (If yes, it requires a JEP). Under each question is a relevant example proposal.
+
+- Does the proposal/implementation require PRs across multiple orgs?
+  - Defining a unique cell identifier
+- Does the proposal/implementation PR impact multiple orgs, or have widespread community impact?
+  - Updating nbformat
+- Does the proposal/implementation change an invariant in one or more orgs?
```
rpwagner, line 146: What does the term `org` mean in this context? Is it referring to a software component, like `ipykernel`, a Jupyter Subproject, or a GitHub org? The intended meaning seems to be software component.

JohanMabille: I think it refers to a GitHub orgs (which are usually related to subprojects, or parts of subprojects). My understanding is that if a changes impacts more than a repo and its direct "dependencies", or "related", then it should probably be a JEP

jep-process-v2/jep-process-v2.md
JohanMabille: > The example of creating a single repo within an existing GitHub org feels too specific for a JEP. However, the creation of a new GitHub org does need a more well-defined process to ensure that it has the appropriate ownership and configuration (e.g., 2FA). I suggest changing this line to simply Creating a new official GitHub organization.

I agree, and the way we have done in recent years tends to confirm that.

> Thinking further, it may be worth a JEP to clarify the creation and management of GitHub entities (orgs, repos, teams, etc.). I think the SSC is the place to handle that JEP, with additional review by the Governance Council.

Totally agree on this! However, I think it should not block this JEP, we can amend it later when the JEP specifying the creation and management of GitHub entities is done.

rpwagner: > Totally agree on this! However, I think it should not block this JEP, we can amend it later when the JEP specifying the creation and management of GitHub entities is done.

Absolutely should not block this one! I was just thinking out loud (or in writing).

jep-process-v2/jep-process-v2.md
rpwagner: Is this clearer about the intention of JEPs, without being specifically about GitHub repos and orgs?

```
- Does the proposal/implementation change an invariant in a separate Jupyter Subproject?
- Does the proposal/implementation impact the interoperability of software components developed in separate Jupyter Subprojects?
```

JohanMabille: Indeed, I find that formulation more precise.

ivanov:
I see a bunch of checkmarks already, @JohanMabille are we moving onto the voting period? if so, let's ping the SSC.

JohanMabille:
@ivanov indeed, I think all the concerns have been addressed. I thought editing the first comment for calling for a vote would ping everyone, it didn't work?

ivanov:
> I thought editing the first comment for calling for a vote would ping everyone, it didn't work?

indeed it did not (I was just notified of changes here from prior participation)

@jupyter/software-steering-council - this proposal has moved onto the voting phase.

SylvainCorlay (approved):

PR #105, opened 2023-04-19
Kernelspec JSON schema

JohanMabille:
# Kernelspec JSON schema

This proposes to specify the kernelspec through a JSON schema, and have the specification and documentation live in a [dedicated repo](https://github.com/jupyter-standards/kernel-protocol).

The JSON schema is joined as a standalone file.

Like #106 , let's move to the voting phase. As said in [this comment](https://github.com/jupyter/enhancement-proposals/pull/105#issuecomment-1965239417), future schema should not be JEPs, but PR open to the [Jupyter Schema repo](https://github.com/jupyter/schema)

- @gabalafou
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @ivanov
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @JohanMabille 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @jtpio 
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @marthacryan
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @martinRenou
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @minrk
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @rpwagner 
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @SylvainCorlay
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @vidartf 
  - [ ] Yes
  - [ ] No 
  - [ ] Abstain

### DIFF
--- a/kernelspec-spec/kernelspec-spec.md
+++ b/kernelspec-spec/kernelspec-spec.md
@@ -2,7 +2,7 @@
 title: kernelspec specification
 authors: Johan Mabille
 issue-number: XX
-pr-number: XX
+pr-number: [#105](https://github.com/jupyter/enhancement-proposals/pull/105)
 date-started: "2023-04-19"
 ---
 
--- a/kernelspec-spec/kernelspec-spec.md
+++ b/kernelspec-spec/kernelspec-spec.md
@@ -1,0 +1,29 @@
+---
+title: kernelspec specification
+authors: Johan Mabille
+issue-number: XX
+pr-number: XX
+date-started: "2023-04-19"
+---
+
+# Specification of the kernelspec
+
+## Problem
+
+The kernelspec configuration file is documented aside the kernel protocol documentation, but it is not
+*specified*. Besides, it might be unclear whether the kernelspec is part of the kernel protocol, or
+independent.
+
+## Proposed Enhancement
+
+We propose to specify the kernelspec with the JSON schema joined in this PR. The specification is conform
+to [the current description of the kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs),
+and adds an optional `protocol_version` field.
+
+[A dedicated repo](https://github.com/jupyter-standards/kernelspec) for the specification and the documentation of
+the kernelspec has been created.
+
+### Impact on existing implementations
+
+None, this JEP only adds an optional field in the kernelspec.
+
--- a/kernelspec-spec/kernelspec-spec.md
+++ b/kernelspec-spec/kernelspec-spec.md
@@ -10,18 +10,15 @@ date-started: "2023-04-19"
 
 ## Problem
 
-The kernelspec configuration file is documented aside the kernel protocol documentation, but it is not
-*specified*. Besides, it might be unclear whether the kernelspec is part of the kernel protocol, or
-independent.
+The kernelspec configuration file is [documented](https://github.com/jupyter/jupyter_client/blob/main/docs/kernels.rst) aside the kernel protocol documentation, and an [implementation](https://github.com/jupyter/jupyter_client/blob/main/jupyter_client/kernelspec.py#L21) is available, but it is not *specified*. Besides, it might be unclear whether the kernelspec is part of the kernel protocol, or independent.
 
 ## Proposed Enhancement
 
-We propose to specify the kernelspec with the JSON schema joined in this PR. The specification is conform
-to [the current description of the kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs),
-and adds an optional `protocol_version` field.
+We propose to specify the kernelspec with the JSON schema joined in this PR. The specification would reflect
+[the current description of the kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs),
+and adds an optional `kernel_protocol_version` field.
 
-[A dedicated repo](https://github.com/jupyter-standards/kernelspec) for the specification and the documentation of
-the kernelspec has been created.
+The specification and the documentation of the kernelspec will be stored aside [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol)
 
 ### Impact on existing implementations
 
--- a/kernelspec-spec/kernelspec-spec.md
+++ b/kernelspec-spec/kernelspec-spec.md
@@ -18,7 +18,7 @@ We propose to specify the kernelspec with the JSON schema joined in this PR. The
 [the current description of the kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs),
 and adds an optional `kernel_protocol_version` field.
 
-The documentation of the kernelspec will be stored aside [that of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol). The schema will be stored in a dedicated repo for all Jupyter schemas.
+The documentation of the kernelspec will be stored aside [that of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol). The schema will be stored in the [dedicated repo for all Jupyter schemas](https://github.com/jupyter/schema).
 
 ### Impact on existing implementations
 
--- a/kernelspec-spec/kernelspec-spec.md
+++ b/kernelspec-spec/kernelspec-spec.md
@@ -18,7 +18,7 @@ We propose to specify the kernelspec with the JSON schema joined in this PR. The
 [the current description of the kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs),
 and adds an optional `kernel_protocol_version` field.
 
-The specification and the documentation of the kernelspec will be stored aside [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol)
+The documentation of the kernelspec will be stored aside [that of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol). The schema will be stored in a dedicated repo for all Jupyter schemas.
 
 ### Impact on existing implementations
 
--- a/kernelspec-spec/kernelspec.schema.json
+++ b/kernelspec-spec/kernelspec.schema.json
@@ -6,7 +6,7 @@
     "type": "object",
 
     "definitions": {
-        "argv": {
+        "kernel_arguments": {
             "type": "object",
             "required": ["argv"],
             "properties": {
--- a/kernelspec-spec/kernelspec.schema.json
+++ b/kernelspec-spec/kernelspec.schema.json
@@ -1,0 +1,44 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://standards.jupyter.org/kernelspec.schema.json",
+    "title": "Kernelspec",
+    "description": "Specification of the kernelspec file",
+    "type": "object",
+    "properties": {
+        "argv": {
+            "description": "A list of command line arguments used to start the kernel. The text {connection_file} in any argument will be replaced with the path to the connection file.",
+            "type": "array",
+            "items": {
+                "type": "string"
+            },
+            "minItems": 3
+        },
+        "display_name": {
+            "description": "The kernel’s name as it should be displayed in the UI. Unlike the kernel name used in the API, this can contain arbitrary unicode characters.",
+            "type": "string"
+        },
+        "language": {
+            "description": "The name of the language of the kernel. When loading notebooks, if no matching kernelspec key (may differ across machines) is found, a kernel with a matching language will be used. This allows a notebook written on any Python or Julia kernel to be properly associated with the user’s Python or Julia kernel, even if they aren’t listed under the same name as the author’s.",
+            "type": "string"
+        },
+        "kernel_protocol_version": {
+            "description": "The version of protocol this kernel implements. If not specified, the client will assume the version is <5.5 until it can get it via the kernel_info request.",
+            "type": "string"
+        },
+        "interrupt_mode": {
+            "description": "May be either signal or message and specifies how a client is supposed to interrupt cell execution on this kernel, either by sending an interrupt signal via the operating system’s signalling facilities (e.g. SIGINT on POSIX systems), or by sending an interrupt_request message on the control channel (see Kernel interrupt). If this is not specified the client will default to signal mode.",
+            "type": "string",
+            "enume": ["signal", "message"]
+        },
+        "env": {
+            "description": "A dictionary of environment variables to set for the kernel. These will be added to the current environment variables before the kernel is started. Existing environment variables can be referenced using ${<ENV_VAR>} and will be substituted with the corresponding value. Administrators should note that use of ${<ENV_VAR>} can expose sensitive variables and should use only in controlled circumstances.",
+            "type": "object",
+            "additionalProperties": {"type": "string" }
+        },
+        "metadata": {
+            "description": "A dictionary of additional attributes about this kernel; used by clients to aid in kernel selection. Metadata added here should be namespaced for the tool reading and writing that metadata.",
+            "type": "object"
+        }
+    },
+    "required": ["argv", "display_name", "language"]
+}
--- a/kernelspec-spec/kernelspec.schema.json
+++ b/kernelspec-spec/kernelspec.schema.json
@@ -16,7 +16,7 @@
                     "items": {
                         "type": "string"
                     },
-                    "minItems": 3
+                    "minItems": 1
                 }
             }
         },
--- a/kernelspec-spec/kernelspec.schema.json
+++ b/kernelspec-spec/kernelspec.schema.json
@@ -47,7 +47,7 @@
                 "kernel_protocol_version": {
                     "description": "The version of protocol this kernel implements. If not specified, the client will assume the version is <5.5 until it can get it via the kernel_info request. The kernel protocol uses semantic versioning (SemVer).",
                     "type": "string",
-                    "pattern": "^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
+                    "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
                     }
             }
         },
--- a/kernelspec-spec/kernelspec.schema.json
+++ b/kernelspec-spec/kernelspec.schema.json
@@ -1,47 +1,98 @@
 {
     "$schema": "https://json-schema.org/draft/2020-12/schema",
-    "$id": "https://schemas.jupyter.org/kernelspec-v1.0.schema.json",
+    "$id": "https://schema.jupyter.org/kernelspec-v1.0.schema.json",
     "title": "Jupyter Kernelspec",
     "description": "A description of the data required to start and manage a Jupyter Kernel",
     "type": "object",
 
-    "properties": {
+    "definitions": {
         "argv": {
-            "description": "A list of command line arguments used to start the kernel. The text {connection_file} in any argument will be replaced with the path to the connection file.",
-            "type": "array",
-            "items": {
-                "type": "string"
-            },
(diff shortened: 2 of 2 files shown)


kernelspec-spec/kernelspec-spec.md
```diff
@@ -0,0 +1,29 @@
...
+issue-number: XX
+pr-number: XX
+date-started: "2023-04-19"
+---
+
+# Specification of the kernelspec
+
+## Problem
+
+The kernelspec configuration file is documented aside the kernel protocol documentation, but it is not
```
bollwyvl, line 13: perhaps add a [link](https://github.com/jupyter/jupyter_client/blob/main/docs/kernels.rst), and to the [implementation](https://github.com/jupyter/jupyter_client/blob/main/jupyter_client/kernelspec.py#L21).

The scope of this should likely include the connection file as well?

kernelspec-spec/kernelspec-spec.md
```diff
@@ -0,0 +1,29 @@
...
+*specified*. Besides, it might be unclear whether the kernelspec is part of the kernel protocol, or
+independent.
+
+## Proposed Enhancement
+
+We propose to specify the kernelspec with the JSON schema joined in this PR. The specification is conform
```
bollwyvl, line 19: As it _isn't_ a spec yet, perhaps "specification would reflect"

kernelspec-spec/kernelspec.schema.json
```diff
@@ -0,0 +1,44 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://standards.jupyter.org/kernelspec.schema.json",
+    "title": "Kernelspec",
+    "description": "Specification of the kernelspec file",
+    "type": "object",
+    "properties": {
```
bollwyvl, line 7: recommend some top-level `definitions` to make individual pieces easier to reference in other schema.

kernelspec-spec/kernelspec.schema.json
```diff
@@ -0,0 +1,44 @@
...
+            "enume": ["signal", "message"]
+        },
+        "env": {
+            "description": "A dictionary of environment variables to set for the kernel. These will be added to the current environment variables before the kernel is started. Existing environment variables can be referenced using ${<ENV_VAR>} and will be substituted with the corresponding value. Administrators should note that use of ${<ENV_VAR>} can expose sensitive variables and should use only in controlled circumstances.",
+            "type": "object",
+            "additionalProperties": {"type": "string" }
+        },
+        "metadata": {
+            "description": "A dictionary of additional attributes about this kernel; used by clients to aid in kernel selection. Metadata added here should be namespaced for the tool reading and writing that metadata.",
+            "type": "object"
```
bollwyvl, line 40: As this is not widely-used yet, could this _already_ be further constrained to enforce namespacing:

```toml
[definitions.metadata]
type = "object"

[definitions.metadata.additionalProperties]
type = "object"
```

kernelspec-spec/kernelspec.schema.json
```diff
@@ -0,0 +1,44 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://standards.jupyter.org/kernelspec.schema.json",
```
bollwyvl, line 3: we aren't (and probably shouldn't) claim that we are a standards body at the moment... `schemas` or `specs` is probably more accurate.

bollwyvl: also, this `$id` should probably reflect a version

kernelspec-spec/kernelspec-spec.md
```diff
@@ -0,0 +1,29 @@
...
+to [the current description of the kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs),
+and adds an optional `protocol_version` field.
+
+[A dedicated repo](https://github.com/jupyter-standards/kernelspec) for the specification and the documentation of
+the kernelspec has been created.
```
krassowski, line 24: In the Notebook format, precisely nbformat v4, in metadata there is a `kernelspec` property, with schema defined [here](https://github.com/jupyter/nbformat/blob/633a433d13436247c5994f571911c74a1ec507d4/nbformat/v4/nbformat.v4.5.schema.json#L13-L27). Previously (in nbformat v3) this was named as `kernel_info` (side note: [format description is outdated](https://nbformat.readthedocs.io/en/latest/format_description.html#top-level-structure) as it still mentions it).

Should the JEP also give mandate to update nbformat schema so that it refer to this new kernelspec a source of truth (using definition link)? In `nbformat` v4 `language` is not required.

Can we describe advantages of moving kernelspec schema out of nbformat in the JEP to explain the reasoning?

kernelspec-spec/kernelspec-spec.md
JohanMabille: > perhaps add a [link](https://github.com/jupyter/jupyter_client/blob/main/docs/kernels.rst), and to the [implementation](https://github.com/jupyter/jupyter_client/blob/main/jupyter_client/kernelspec.py#L21).

Thanks for the suggestion, I'll update the JEP.

> The scope of this should likely include the connection file as well?

That is something I was wondering when writing the schema. I don't have a strong opinion on this, I just wanted to have the minimal spec that could be quickly approved since other JEPs depend on it.

kernelspec-spec/kernelspec.schema.json
```diff
@@ -0,0 +1,44 @@
...
+    "title": "Kernelspec",
```
bollwyvl, line 4: Recommend a more verbose title, perhaps "Jupyter Kernelspec"

kernelspec-spec/kernelspec.schema.json
```diff
@@ -0,0 +1,44 @@
...
+            "type": "string"
+        },
+        "kernel_protocol_version": {
+            "description": "The version of protocol this kernel implements. If not specified, the client will assume the version is <5.5 until it can get it via the kernel_info request.",
+            "type": "string"
+        },
+        "interrupt_mode": {
+            "description": "May be either signal or message and specifies how a client is supposed to interrupt cell execution on this kernel, either by sending an interrupt signal via the operating system’s signalling facilities (e.g. SIGINT on POSIX systems), or by sending an interrupt_request message on the control channel (see Kernel interrupt). If this is not specified the client will default to signal mode.",
+            "type": "string",
+            "enume": ["signal", "message"]
```
bollwyvl, line 31: `enum`

kernelspec-spec/kernelspec-spec.md
```diff
@@ -0,0 +1,29 @@
...
+*specified*. Besides, it might be unclear whether the kernelspec is part of the kernel protocol, or
+independent.
+
+## Proposed Enhancement
+
+We propose to specify the kernelspec with the JSON schema joined in this PR. The specification is conform
+to [the current description of the kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs),
+and adds an optional `protocol_version` field.
+
+[A dedicated repo](https://github.com/jupyter-standards/kernelspec) for the specification and the documentation of
```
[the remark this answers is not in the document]
JohanMabille: I agree that a repo-per-spec might be overkill. However, I would like to have some separation of concerns. The repo for the kernel protocol will also hold the roadmap and additional documentation about the "release" process of the protocol. Also the kernel protocol and the widget protocol are somehow orthogonal, I would find it confusing to have them in the same repo.

kernelspec-spec/kernelspec.schema.json
Zsailer: I emailed the EC today about creating a domain for hosting such schemas... I was thinking `schemas.jupyter.org`.

> also, this $id should probably reflect a version
 
:+1:

kernelspec-spec/kernelspec.schema.json
```diff
@@ -0,0 +1,44 @@
...
+            "description": "The kernel’s name as it should be displayed in the UI. Unlike the kernel name used in the API, this can contain arbitrary unicode characters.",
+            "type": "string"
+        },
+        "language": {
+            "description": "The name of the language of the kernel. When loading notebooks, if no matching kernelspec key (may differ across machines) is found, a kernel with a matching language will be used. This allows a notebook written on any Python or Julia kernel to be properly associated with the user’s Python or Julia kernel, even if they aren’t listed under the same name as the author’s.",
+            "type": "string"
+        },
+        "kernel_protocol_version": {
+            "description": "The version of protocol this kernel implements. If not specified, the client will assume the version is <5.5 until it can get it via the kernel_info request.",
+            "type": "string"
```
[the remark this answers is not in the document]
Zsailer: I don't think this should be included in the first version of the kernelspec. 

I believe the first _published_ version should only include the spec currently documented in the protocol documentation. 

Then, we can quickly publish a second version with kernel_protocol_version added as a key. That way, we don't immediately invalidate everyone kernelspecs. 

Also, creates a circular dependency on JEP #66.

kernelspec-spec/kernelspec.schema.json
JohanMabille: I'm fine with both names, let me know which one is created and I will update the JEP accordingly.

kernelspec-spec/kernelspec.schema.json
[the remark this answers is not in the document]
Zsailer: > That way, we don't immediately invalidate everyone kernelspecs.

I realize this actually isn't true, since "kernel_protocol_version" wouldn't be a required field. That said, we still would need to settle the naming in #66 to merge this. I think "kernel_protocol_version" is a fine name for the field, so maybe this isn't any issue.

Zsailer:
Thank you for doing this @JohanMabille! You beat me to it. I had spent some time yesterday writing the exact same JEP as an action item from the SSC meeting. 😅

JohanMabille:
@Zsailer arf sorry for that!  This came out when discussing the Handshake pattern, we need an additional field in the kernelspec and I thought it could be the opportunity to add a schema for it. I'll be more careful next time to avoid doing the work twice!

Zsailer:
> @Zsailer arf sorry for that! This came out when discussing the Handshake pattern, we need an additional field in the kernelspec and I thought it could be the opportunity to add a schema for it. I'll be more careful next time to avoid doing the work twice!

No no no—this is fantastic! I really appreciate you taking this over! No need to apologize 😃

kernelspec-spec/kernelspec-spec.md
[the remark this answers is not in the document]
Zsailer: > I'm not sold yet on repo-per-spec, and think the "big tent" repo, with common tooling and interlinked documentation will be more sustainable.

I would agree with this. I think it would be more sustainable and easier for folks to follow if we had a single "Jupyter specs" repo.

kernelspec-spec/kernelspec-spec.md
Zsailer: I think the "kernelspec" entry in nbformat.v4 is a misnomer. It is not a kernelspec—sure, it shares some of the same fields, but there is a subtle difference here. 

The kernelspec file does not  have a "name" field. The (locally) unique name for a kernelspec comes from the directory on disk where the kernel.json file is located—it is not stored in the kernelspec file. 

In the nbformat, this "name" field is necessary to locate the kernelspec on disk (this has always been brittle, because it means this information becomes useless the moment the notebook leaves the current context).

It really should be called something like "kernelspec_info". 

> Should the JEP also give mandate to update nbformat schema so that it refer to this new kernelspec a source of truth (using definition link)? In nbformat v4 language is not required.

This would be great for notebooks that always run in the same Python/conda/virtual environment (this field isn't helpful for notebooks that move around, but that's a separate issue). I would be in favor of changing this field, but I think this is out-of-scope for this JEP. We should make a follow-up JEP to update nbformat if we'd like to pursue that. 

> Can we describe advantages of moving kernelspec schema out of nbformat in the JEP to explain the reasoning?
>
>> The kernelspec is really about describing a kernel, and it is totally orthogonal to the notebook format. Kernel authors should not have to refer to the ntebook format spec to know how to implement a kernelspec.

I believe this is out-of-scope for the current JEP. 

That said, they aren't entirely orthogonal. Yes, the kernelspec shouldn't depend on the nbformat, but the nbformat _loosely_ depends on the kernelspec today. I think that's what @krassowski  is alluding to... if we can drop nbformat's current conflicting definition of the kernelspec and point at this definition, we gain some standardization.

kernelspec-spec/kernelspec-spec.md
```diff
@@ -0,0 +1,29 @@
...
+
+The kernelspec configuration file is documented aside the kernel protocol documentation, but it is not
+*specified*. Besides, it might be unclear whether the kernelspec is part of the kernel protocol, or
+independent.
+
+## Proposed Enhancement
+
+We propose to specify the kernelspec with the JSON schema joined in this PR. The specification is conform
+to [the current description of the kernelspec](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs),
+and adds an optional `protocol_version` field.
```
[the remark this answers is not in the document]
JohanMabille: Not sure to get you, do you mean a schema for describing the version? Or having a schema as a value for the field?

Zsailer:
I think it's worth saying for voting purposes—I'm :+1: for this JEP getting accepted.

We can continue to copy-edit, but wanted to explicitly submit my support. 😃

kernelspec-spec/kernelspec.schema.json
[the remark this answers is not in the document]
JohanMabille: Actually let's say that #66 depends on this one and would be updated (if needed) when this one is accepted.

JohanMabille:
More changes to come, but I need to educate myself more to JSON schemas :sweat_smile:

kernelspec-spec/kernelspec.schema.json
JohanMabille: I don't find a way of grouping pieces, they all appear independent to me, having a definition per field would just add complexity and not ease their reference in other shcema, right? Sorry for the naive question, but I'm not used to JSON schemas.

bollwyvl: The kernelspec, and specific fields within it, need to be referenced by other schema.

Have a look at, e.g. [nbformat](https://github.com/jupyter/nbformat/blob/main/nbformat/v4/nbformat.v4.5.schema.json#L13). If the kernelspec is formalized, then conventionally specs would like to point to specific _meanings_ without pointing at a concrete _locations_:

```
$ref = "https://schema.jupyter.org/kernelspec/v1/schema.json#/definitions/display-name"
```
vs

```
$ref = "https://schema.jupyter.org/kernelspec/v1/schema.json#/properties/display_name"
```

This becomes more relevant, once more complex patterns are used such as `additionalProperties` or `patternProperties` (useful for e.g. mimetypes).

kernelspec-spec/kernelspec-spec.md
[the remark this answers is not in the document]
bollwyvl: See https://github.com/jupyter/enhancement-proposals/pull/97 for some of the motivation.

In a nutshell: having a single, top-level `$schema` const is more portable and inspectable than number-based values in custom fields.

kernelspec-spec/kernelspec-spec.md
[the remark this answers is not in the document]
bollwyvl: Sure, widgets are _maybe_ an exception... but the knowledge of how `comm` works is not. At the end of the chain, something like the [Jupyter Server API](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/api/api.yaml) references _almost_ everything each other... kernelspec just happens to be very low in the order, with only perhaps a `known-mimetypes.json` below it, as that is _not_ a schema-level construct like `uri`. 

When changes occur, often these things will reference each other, and documentation and generated type packages should reflect the compatibility of them.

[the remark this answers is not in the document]
JohanMabille: How comm works is part of the Kernel protocol, thus I agree the Widgets protocol references the Kernel protocol, but not vice versa. And I took the Widget protocol as an example, but we can consider other things, like the LSP, which is totally unrelated. I see this as a software distribution, we have low layers upon which are built upper layers, and there should not be circular dependencies. The Jupyter Server API you mentioned is a super high layer that references everything, however lower layers should not reference it.

kernelspec-spec/kernelspec.schema.json
JohanMabille: Thanks for the detailed explanation, I will try to make the appropriate changes.

kernelspec-spec/kernelspec-spec.md
JohanMabille: I have opened a dedicated JEP for the connection file, see https://github.com/jupyter/enhancement-proposals/pull/106

kernelspec-spec/kernelspec.schema.json
```diff
@@ -0,0 +1,98 @@
...
+            }
+        },
+        "kernel_protocol_version": {
+            "type": "object",
+            "required": ["kernel_protocol_version"],
+            "properties": {
+                "kernel_protocol_version": {
+                    "description": "The version of protocol this kernel implements. If not specified, the client will assume the version is <5.5 until it can get it via the kernel_info request. The kernel protocol uses semantic versioning (SemVer).",
+                    "type": "string",
+                    "pattern": "^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
```
fcollonval, line 50:
```suggestion
                    "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
```

westurner:
- [ ] BLD,SCH: CI task to run e.g. js2shacl to generate W3C SHACL from the
kernelspec JSON schema

https://github.com/ThiagoCComelli/JS2SHACL-JSON-Schema-to-SHACL-conversor/blob/master/conversor.js

IDK why SHACL also yet, but there's a converter

kernelspec-spec/kernelspec-spec.md
```diff
@@ -0,0 +1,26 @@
+---
+title: kernelspec specification
+authors: Johan Mabille
+issue-number: XX
+pr-number: XX
```
jasongrout, line 5:
```suggestion
pr-number: #105
```

kernelspec-spec/kernelspec.schema.json
ianthomas23: The extra backslashes are needed for the JSON file to be readable.

But I don't think need to get the schema completely correct here, the canonical schema will live in the [schema repo](https://github.com/jupyter/schema) and eventually be uploaded to https://schema.jupyter.org. The version here will form the basis for that but it will no doubt change when we have written a few schemas and worked out a standard approach for relationships between schemas for example.

JohanMabille: I applied the change so that the JSON is readable; I agree that we don't need to have the final version of the schema in this JEP as long as we have something correct and usable. We can then fastly iterate in the schema repo to robustify it.

JohanMabille:
Kind ping on this, I think this is ready for a final review:
- The [Jupyter Schema repo](https://github.com/jupyter/schema) has been created, it will hold all the schemas for Project Jupyter and the JEP has been updated accodingly (not more specific repo per spec)
- While the first schema specifications have been proposed through JEPs, the recent works and meetings of the Jupyter Contributors involved in the schema specifications led to the conclusion that we could directly work on the Jupyter Schema repo: implement first versions of different schema and then have fast iterations to make them more robust and extract common parts into dedicated schema. This would streamline the process.

kernelspec-spec/kernelspec.schema.json
```diff
@@ -0,0 +1,98 @@
...
+            "type": "object",
+            "required": ["argv"],
+            "properties": {
+                "argv": {
+                    "description": "A list of command line arguments used to start the kernel. The text {connection_file} in any argument will be replaced with the path to the connection file.",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "minItems": 3
```
minrk, line 19: I don't believe this needs to be 3, e.g. `['kernel-executable', '{connection_file}']`. I _think_ the connection file is allowed to be passed via the env and not the CLI (not certain, but this should probably be possible in the future), so I think the real minimum length is 1:

```suggestion
                    "minItems": 1
```

minrk (approved):
Made one small suggestion on minItems, but 👍

JohanMabille:
The JEP has been approved.

Yes: 9
No: 0
Abstention: 0
No vote: 1

Merging

PR #106, opened 2023-04-22
JEP for specifying the connectionfile

JohanMabille:
# Connection file JSON schema

This proposes to specify the connection file through a JSON schema, and have the specification and documentation live in a [dedicated repo](https://github.com/jupyter-standards/kernel-protocol).

The JSON schema is joined as a standalone file.

---

I(s been a while since we proposed to move to the voting phase, let's fix it now ;) Ntice that further JSON specs should not been JEPs, but PRs to https://github.com/jupyter/schema

- @gabalafou
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @ivanov
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @JohanMabille 
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @jtpio 
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @marthacryan
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @martinRenou
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @minrk
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @rpwagner 
  - [x] Yes
  - [ ] No
  - [ ] Abstain 
- @SylvainCorlay
  - [x] Yes
  - [ ] No
  - [ ] Abstain
- @vidartf 
  - [x] Yes
  - [ ] No 
  - [ ] Abstain

### DIFF
--- a/connectionfile-spec/connectionfile-spec.md
+++ b/connectionfile-spec/connectionfile-spec.md
@@ -15,10 +15,9 @@ The connection file is [documented](https://github.com/jupyter/jupyter_client/bl
 ## Proposed Enhancement
 
 We propose to specify the connection file with the JSON schema joined in this PR. The specification would reflect
-[the current description of the connection file](https://jupyter-client.readthedocs.io/en/stable/kernels.html#connection-files),
-and adds an optional `kernel_protocol_version` field.
+[the current description of the connection file](https://jupyter-client.readthedocs.io/en/stable/kernels.html#connection-files).
 
-The specification and the documentation of the connection file will be stored along side [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol)
+The documentation of the connection file will be stored along side [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol) while its specification will be stored in the [Jupyter schema repo](https://github.com/jupyter/schema).
 
 ### Impact on existing implementations
 
--- a/connectionfile-spec/connectionfile-spec.md
+++ b/connectionfile-spec/connectionfile-spec.md
@@ -1,0 +1,25 @@
+---
+title: connection file specification
+authors: Johan Mabille
+issue-number: XX
+pr-number: XX
+date-started: "2023-04-19"
+---
+
+# Specification of the connection file
+
+## Problem
+
+The connection file is [documented](https://github.com/jupyter/jupyter_client/blob/main/docs/kernels.rst) aside the kernel protocol documentation, but it is not *specified*.
+
+## Proposed Enhancement
+
+We propose to specify the connection file with the JSON schema joined in this PR. The specification would reflect
+[the current description of the connection file](https://jupyter-client.readthedocs.io/en/stable/kernels.html#connection-files),
+and adds an optional `kernel_protocol_version` field.
+
+The specification and the documentation of the connectio filewill be stored aside [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol)
+
+### Impact on existing implementations
+
+None, this JEP only specifies the current implementations.
--- a/connectionfile-spec/connectionfile-spec.md
+++ b/connectionfile-spec/connectionfile-spec.md
@@ -18,7 +18,7 @@ We propose to specify the connection file with the JSON schema joined in this PR
 [the current description of the connection file](https://jupyter-client.readthedocs.io/en/stable/kernels.html#connection-files),
 and adds an optional `kernel_protocol_version` field.
 
-The specification and the documentation of the connectio filewill be stored aside [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol)
+The specification and the documentation of the connection file will be stored along side [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol)
 
 ### Impact on existing implementations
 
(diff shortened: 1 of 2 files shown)


connectionfile-spec/connectionfile.schema.json
```diff
@@ -0,0 +1,64 @@
...
+    "type": "object",
+
+    "definitions": {
+        "signature": {
+            "type": "object",
+            "required": ["signature_scheme", "key"],
+            "properties": {
+                "signature_scheme": {
+                    "type": "string",
+                    "description": "scheme used to sign the messages"
```
minrk, line 15: Should this be an enum and require revisions to support new schemes? Or merely include some example values in the description (i.e. it's ~always hmac-256)?

connectionfile-spec/connectionfile.schema.json
```diff
@@ -0,0 +1,64 @@
...
+                    "type": "string",
+                    "description": "transport protocol",
+                    "enum": ["tcp", "ipc", "inproc"]
+                },
+                "ip": {
+                    "type": "string",
+                    "description": "ip of the machine where the kernel runs"
+                },
+                "shell_port": {
+                    "type": "string",
```
minrk, line 37: Ports are generally integers with tcp, but may be strings when using ipc.

connectionfile-spec/connectionfile-spec.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Problem
+
+The connection file is [documented](https://github.com/jupyter/jupyter_client/blob/main/docs/kernels.rst) aside the kernel protocol documentation, but it is not *specified*.
+
+## Proposed Enhancement
+
+We propose to specify the connection file with the JSON schema joined in this PR. The specification would reflect
+[the current description of the connection file](https://jupyter-client.readthedocs.io/en/stable/kernels.html#connection-files),
+and adds an optional `kernel_protocol_version` field.
```
minrk, line 19: the kernel_protocol_version field is not in the proposed spec.

connectionfile-spec/connectionfile-spec.md
```diff
@@ -0,0 +1,25 @@
...
+
+The specification and the documentation of the connectio filewill be stored aside [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol)
```
minrk, line 21:
```suggestion
The specification and the documentation of the connection file will be stored along side [those of the kernel protocol](https://github.com/jupyter-standards/kernel-protocol)
```

connectionfile-spec/connectionfile.schema.json
JohanMabille: I agree that an enum would be better, I will amend the schema accordingly.

JohanMabille:
@jupyter/software-steering-council kind ping on this one, similar to  https://github.com/jupyter/enhancement-proposals/pull/105

Zsailer:
I propose we move this to a voting phase. This has been open for nearly a year, so I think we can safely close the comment period. Now that the jupyter/schema repo is moving forward, this seems like a good time to proceed.

connectionfile-spec/connectionfile.schema.json
```diff
@@ -0,0 +1,64 @@
...
+                    "description": "transport protocol",
+                    "enum": ["tcp", "ipc", "inproc"]
+                },
+                "ip": {
+                    "type": "string",
+                    "description": "ip of the machine where the kernel runs"
+                },
+                "shell_port": {
+                    "type": ["integer","string"],
```
minrk, line 37: Is it worth using oneOf to distinguish tcp with integer ports from ipc/inproc where it's really a string suffix (that I believe implementation-wise accepts integers)? Essentially, signal that tcp doesn't accept strings for ports.

Fine if that's more schema complexity that folks want.

minrk (approved):

connectionfile-spec/connectionfile.schema.json
```diff
@@ -0,0 +1,64 @@
...
+                },
+                "key": {
+                    "type": "string",
+                    "description": "key used to sign the messages"
+                }
+            }
+        },
+        "kernel_network": {
+            "type": "object",
+            "required": ["transport", "ip", "shell_port", "control_port", "stdin_port", "hb_port", "iopub_port"],
```
vidartf, line 25: I realize I'm late to this discussion, but I just want to understand it fully before voting: Is the spec as proposed now consistent with https://github.com/jupyter/enhancement-proposals/pull/66/files ? If not, how would the schema evolve to support that? Would these fields no longer be required?

ivanov (approved):

SylvainCorlay (approved):

connectionfile-spec/connectionfile.schema.json
SylvainCorlay: My understanding is that both #105 and #106 represent the current state of thing, and that the implementation of #66 would require updating these specs.

vidartf: Ok. My main point was to do a sanity check on whether#66  would require any changes that the spec as it currently is would make harder to implement. It sounds like that isn't a concern, so I voted for.

SylvainCorlay:
The JEP has been approved unanimously.

- Yes: 10
- No: 0
- Abstention: 0

**Merging**

PR #114, opened 2023-06-15
Added missing header to the xpub JEP

JohanMabille:
This is not a JEP, but a technical update of an already merged JEP.


Zsailer (approved):

Zsailer:
Thanks!

PR #118, opened 2023-09-01
Clarify restart to mean restart in-place

### LINKED ISSUE #117: Pre-Proposal: Add Restart In Place API Support
## Problem

Restart-in-place is a new type of restart [requested ](https://github.com/jupyter-server/jupyter_server/issues/592) by the Jupyter community. Normally, when a remote kernel is restarted, both the kernel process and the tunnel process are ended. On resource-managed clusters, ending the tunnel process will end the task that the kernel is scheduled on, forcing the restarted kernel to wait in the back of the queue for another task. This is a major inconvenience for users on distributed clusters who need to restart their kernels often.

With restart-in-place, the kernel process gets terminated but the tunnel process stays alive. Then, the new kernel is launched directly on the existing tunnel process. This may speed up restart times significantly.

There are previously proposed implementations of restart-in-place. All of them were held back by the lack of an officially supported API for restart-in-place, and had to trigger their restarts through non-intuitive methods such as magics. For example, inplace_restarter started a second (nanny) process alongside the kernel and used a magic to tell the nanny process to restart the kernel. However, inplace_restarter could be tricky to setup and was less straightforward to use than a native Jupyterlab button, and there were use cases where a nanny process is undesirable.

## Spec Changes

We propose adding native API support for restart-in-place. We will add functionality to `jupyter-client`'s kernel management infrastructure to distinguish between standard restart and restart-in-place requests. We will also add a new server endpoint on `jupyter-server`: `/api/kernels/{kernel_id}/restart-in-place` to handle restart in place requests. The new endpoint enables users to specify their restart preferences, and allows kernel implementations to handle the restart-in-place.

We also propose adding a new command in `jupyterlab` to call the new restart-in-place API. Then, we will add front-end elements for calling the command such as toolbar buttons and menu bars items.

We prefer this approach to writing a new extension because several classes in `jupyterlab` that manage the kernel lifecycle have private variables. This means that we can't execute all of the restart logic without editing some classes in `jupyterlab`.

## Reference Implementation

We will also add a reference implementation for restart in place using [DistributedProvisioner](https://github.com/jupyter-server/gateway_provisioners/blob/main/gateway_provisioners/distributed.py) in the package `gateway_provisioners`, since we assume that many users of remote kernels on resource managed clusters either use Enterprise Gateway or Gateway Provisioners. Because a kernel provisioner is responsible for managing the kernel and tunnel process, and a gateway provisioner is just an extension of a kernel provisioner for remote services, we can edit the `DistributedProvisioner` to not terminate the tunnel process upon a restart in place.

### Why is this a JEP?

We decided to write a JEP for restart in place because it requires changes to multiple Project Jupyter packages and edits the spec by which `jupyterlab` communicates with `jupyter_server`. Also, while we are pushing one implementation of restart in place for a specific resource manager, we leave open future implementations for other resource managers such as Kubernetes, Spark, Hadoop, etc...

### Will this work with ____ kernel / implementation?

The proposal would support restart-in-place with many different processes. We leave the implementation of restart-in-place up to the specific kernel specification, but regardless of the implementation, we need to a new API endpoint to distinguish whether the user wants a normal restart or a restart in place.

### What if a kernel doesn't support restart-in-place?

If restart-in-place is not supported, the boolean argument that indicates whether a restart is in-place or not is ignored and the kernel restarts as normal.

### Will the frontend be able to tell when restart in place is enabled and active features accordingly?

Yes, we would like to standardize a variable in the kernel specification that will enable restart-in-place functionality. Then,` jupyterlab` would read in the kernel specification to tell whether restart-in-place is enabled or not.

### What are the implementation details of the new API endpoint?

We propose adding an optional boolean keyword argument to the kernel restart methods in [MultiKernelManager ](https://github.com/jupyter/jupyter_client/blob/main/jupyter_client/multikernelmanager.py)and [KernelManager ](https://github.com/jupyter/jupyter_client/blob/main/jupyter_client/manager.py)in `jupyter_client`. If set to true, they will allow subclasses of `KernelManager` to execute a restart-in-place, if possible.

In jupyter_server, we will edit the [KernelActionHandler ](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/kernels/handlers.py) class to take an additional action "restart-in-place". This will call the restart function as normal, except with the restart in place keyword argument set to true.

mlucool:
Resolves #117

### DIFF
--- a/118-restart-clarification/restart-clarification.md
+++ b/118-restart-clarification/restart-clarification.md
@@ -56,5 +56,5 @@ Only the exact wording changes as proposed in [jupyter_client](https://github.co
 
 # Future possibilities
 
-We would make a service that implements a "hard restart" as discussed in the juptyer-server meeting.
+We would make a service that implements a "hard restart" as discussed in the jupyter-server meeting.
 No one on the meeting had an immediate use case for it.
--- a/118-restart-clarification/restart-clarification.md
+++ b/118-restart-clarification/restart-clarification.md
@@ -25,11 +25,11 @@ believe to be the mental model of users when they click "restart".
 # Guide-level explanation
 
 The [protocol](https://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-shutdown) would describe
-restart as restarting only the kernel process and its subprocess *not* any parent process.
+restart as optimally preserving as many resources outside the kernel as possible (e.g. restarting only the kernel process and its subprocess *not* any parent process).
 
 When the kernel is a toplevel process (e.g. local kernels), there is no change.
 
-When the kernel is not a toplevel process (e.g. when used in Enterprise Gateway), restart means only the kernel restarts. To restart the whole progress group, the stop and start messages must be used. It's up to UIs
+When the kernel is not a toplevel process (e.g. when used in Enterprise Gateway), restart often means only the kernel restarts. To restart the whole progress group, the stop and start messages could be used. It's up to UIs
 for how to display this difference (if any).
 
 # Reference-level explanation
@@ -43,6 +43,8 @@ A new message could be added, as proposed in the pre-JEP. In the [Jupyter Server
 we concluded that is is likely most users want restart to only restart the kernel and not potentially reschedule resources. Therefore, a new message was not the best option.
 For the vast majority of kernels (local subprocesses), this change is a no-op because restart-in-place
 is the same as a hard restart since it is the toplevel kernel process.
+
+For users that want a hard restart, a stop followed by a start continues to be available. While this may be less convenient, a UI can trivially hide this two call process from the user.
 
 # Prior art
 
--- a/118-restart-clarification/restart-clarification.md
+++ b/118-restart-clarification/restart-clarification.md
@@ -1,0 +1,58 @@
+---
+title: Restart Clarification
+authors: Marc Udoff (@mlucool)
+issue-number: [#117](https://github.com/jupyter/enhancement-proposals/issues)
+pr-number: [#118](https://github.com/jupyter/enhancement-proposals/pull/118)
+date-started: 2023-08-23
+---
+
+# Summary
+
+The jupyter client protocol around what "restart" means is ambiguous. Per Jupyter Client's [kernel shutdown documention](https://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-shutdown), the only guidance is:
+>  'restart' : bool # False if final shutdown, or True if shutdown precedes a restart
+
+This has led to a situation where certain jupyter subprojects (e.g. enterprise gateway) have interpreted this to mean a "hard restart", or restart the whole process tree.
+
+We propose to clarify this to always mean "restart-in-place", that is only restart the kernel itself. For
+most usage of Jupyter, this change is a no-op as subprocess kernels already act like this.
+
+# Motivation
+
+This greatly improves the usability of remote kernels whose startup includes scheduling. By making this change,
+restarts for remote kernels will be nearly as fast as those of local kernels. It also matches what we
+believe to be the mental model of users when they click "restart".
+
+# Guide-level explanation
+
+The [protocol](https://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-shutdown) would describe
+restart as restarting only the kernel process and its subprocess *not* any parent process.
+
+When the kernel is a toplevel process (e.g. local kernels), there is no change.
+
+When the kernel is not a toplevel process (e.g. when used in Enterprise Gateway), restart means only the kernel restarts. To restart the whole progress group, the stop and start messages must be used. It's up to UIs
+for how to display this difference (if any).
+
+# Reference-level explanation
+
+The `jupyter-client` messaging page would be updated to indicate this nuance. Any implementations that
+treat restart differently would be updated to match this clarification.
+
+# Rationale and alternatives
+
+A new message could be added, as proposed in the pre-JEP. In the [Jupyter Server meeting](https://github.com/jupyter-server/team-compass/issues/45#issuecomment-1682582186),
+we concluded that is is likely most users want restart to only restart the kernel and not potentially reschedule resources. Therefore, a new message was not the best option.
+For the vast majority of kernels (local subprocesses), this change is a no-op because restart-in-place
+is the same as a hard restart since it is the toplevel kernel process.
+
+# Prior art
+
+N/A
+
+# Unresolved questions
+
+Only the exact wording changes as proposed in [jupyter_client](https://github.com/jupyter/jupyter_client/pull/966).
+
+# Future possibilities
+
+We would make a service that implements a "hard restart" as discussed in the juptyer-server meeting.
+No one on the meeting had an immediate use case for it.


mlucool:
CC @timg512372 @kevin-bates @davidbrochart @minrk @Zsailer @krassowski @Carreau

118-restart-clarification/restart-clarification.md
```diff
@@ -0,0 +1,58 @@
...
+# Motivation
+
+This greatly improves the usability of remote kernels whose startup includes scheduling. By making this change,
+restarts for remote kernels will be nearly as fast as those of local kernels. It also matches what we
+believe to be the mental model of users when they click "restart".
+
+# Guide-level explanation
+
+The [protocol](https://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-shutdown) would describe
+restart as restarting only the kernel process and its subprocess *not* any parent process.
```
minrk, line 28: I would be a little less specific, and _encourage_ restart to preserve as many resources as make sense, up to but not including the kernel process.

For example, restarting a kernel inside a simple container means the kernel can't be PID 1. Running a kernel via `docker exec` so it can be restarted is definitely nice, but I don't think it should be a requirement. Both implementations should be considered valid spec-wise, and are a _deployment configuration_ choice, not a protocol choice.

I wouldn't consider any current 'hard restarts' to be protocol violations, rather less optimized implementations, and multiple provider-specific behaviors governed by config can make perfect sense without being dictated by the spec.

118-restart-clarification/restart-clarification.md
```diff
@@ -0,0 +1,58 @@
...
+
+The `jupyter-client` messaging page would be updated to indicate this nuance. Any implementations that
+treat restart differently would be updated to match this clarification.
+
+# Rationale and alternatives
+
+A new message could be added, as proposed in the pre-JEP. In the [Jupyter Server meeting](https://github.com/jupyter-server/team-compass/issues/45#issuecomment-1682582186),
+we concluded that is is likely most users want restart to only restart the kernel and not potentially reschedule resources. Therefore, a new message was not the best option.
+For the vast majority of kernels (local subprocesses), this change is a no-op because restart-in-place
+is the same as a hard restart since it is the toplevel kernel process.
```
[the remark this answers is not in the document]
mlucool: Fixed

118-restart-clarification/restart-clarification.md
mlucool: LMK what you think of the changes here and in https://github.com/jupyter/jupyter_client/pull/966/files.

Zsailer:
ping @jupyter/software-steering-council for review

minrk (approved):

minrk:
I think this and https://github.com/jupyter/jupyter_client/pull/966 are good as they are. I think it's clear now that this is a clarification and _suggestion_ for meeting user expectations, and not a protocol-level change or compatibility issue.

118-restart-clarification/restart-clarification.md
```diff
@@ -0,0 +1,60 @@
...
+
+N/A
+
+# Unresolved questions
+
+Only the exact wording changes as proposed in [jupyter_client](https://github.com/jupyter/jupyter_client/pull/966).
+
+# Future possibilities
+
+We would make a service that implements a "hard restart" as discussed in the juptyer-server meeting.
```
Zsailer, line 59:
```suggestion
We would make a service that implements a "hard restart" as discussed in the jupyter-server meeting.
```

Zsailer:
Hey @jupyter/software-steering-council 👋 

Since this is a fairly small, straightforward JEP that simply clarifies some expectations around the kernel protocol and discussion has been fairly quite, I propose that we move this into a voting phase by next week. 

If you have any concerns/comments, please raise them immediately. Thanks all!

Zsailer:
Moving this to a voting phase. Pinging the @jupyter/software-steering-council for votes. 

This vote should end on November 13th, 2023.

- @echarles 
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @fcollonval 
  - [ ] Yes
  - [ ] No 
  - [ ] Abstain
- @gabalafou
  - [ ] Yes
  - [ ] No
  - [ ] Abstain 
- @ibdafna 
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @ivanov
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @JohanMabille 
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @minrk
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @rpwagner 
  - [ ] Yes
  - [ ] No
  - [ ] Abstain 
- @SylvainCorlay
  - [ ] Yes
  - [ ] No
  - [ ] Abstain
- @Zsailer 
  - [ ] Yes
  - [ ] No 
  - [ ] Abstain

Zsailer (approved):

118-restart-clarification/restart-clarification.md
echarles: Before casting my vote, is it possible to get an answer on this comment? I find the raised point `A restart should optimally preserve as many resources outside the kernel as possible` important, and I don't see it really adressed in the current text of the JEP.

echarles: @Zsailer I also tried to track the timeline of the comments and @minrk approval... and trying to make sense of the current text vs the PR text opened by @mlucool in the jupyter-client. Globally, I would rather have jupyter-client link to this JEP for the details, rather than having two parallel texts subject to interpretation. Maybe I am looking too much at the details... I am OK if someone closes this comment without further details, but would for sure prefer having the clarifications I am missing in the text rather than in the comments.

rpwagner (approved):

118-restart-clarification/restart-clarification.md
Zsailer: No no, you're making a great case here @echarles. There is something missing from the JEP that's obfuscating what's actually being proposed here. 

This JEP is a bit unique, so maybe I can explain what happened here and we can figure out how explain this in the JEP.

Currently—Jupyter Client's documentation pages hold the "source of truth" specification of the kernel protocol (ideally we move this documentation to a better home, since jupyter_client is a reference implementation; but let's ignore that for now).

@mlucool is proposing a clarification around an ambiguous part of the protocol, e.g. kernel restart. He could simply open a PR against jupyter_client, like he did, to clarify this point in the docs. However, since this is a protocol clarification—i.e. something much more sensitive than a simple change to jupyter_client docs—it called for a JEP.

The JEP is essentially _asking for permission_ to merge the jupyter_client PR and update the protocol with those changes. 

> Globally, I would rather have jupyter-client link to this JEP for the details, rather than having two parallel texts subject to interpretation. 

I agree that having the text in two places is confusing—technically, it belongs _only_ in the jupyter_client PR. Following your logic here, the _JEP_ should reference the jupyter_client PR (not the other way around), but you can understand why that feels awkward as well. Maybe we should make it explicit in the JEP that we're adding language to the protocol specification in jupyter_client and add the link to the draft PR. What do you think, @echarles?

echarles: @Zsailer Thank your for your time to bring explanation and clarification. The kernel protocol specs living in jupyter-client as "source of thuth" is part of the "debt" we need to work on.

> Maybe we should make it explicit in the JEP that we're adding language to the protocol specification in jupyter_client and add the link to the draft PR. What do you think, @echarles?

Agree, so this JEP should just link to @mlucool PR. Once we get enough votes here, we merge both, right?

mlucool: Yes. Did you want me to link back now?

minrk: > maybe @minrk can explicitly comment/approve?

Yes, as indicated by my explicit approval of both PRs after this discussion, I approve it all as it is now.

echarles: Thx for clarification. I have given a positive vote on this PR.

SylvainCorlay (approved):

Zsailer:
Thank you @mlucool for all the time you put into writing this JEP and your patience in the review process. Congratulations on the acceptance! 🚀

mlucool:
Thanks all!

Issue #119, opened 2024-03-26
Pre-proposal: websocket token authentication with subprotocols


minrk:
## Background

Combining these facts:

- Jupyter Servers (often) use API tokens for authentication
- auth tokens are typically passed in the Authorization header
- websockets cannot pass tokens in Authorization headers because the browser implementation [forbids it](https://github.com/whatwg/websockets/issues/16), though most implementations outside the browser do support it.
  As a result, websocket requests must pass tokens in a URL parameter.
- Passing tokens in a URL parameter is generally frowned upon, but not strictly insecure and indeed [explicitly recommended by Browser websocket implementers](https://github.com/whatwg/websockets/issues/16#issuecomment-347180825)

motivates having a new mechanism by which to pass the auth token for websocket requests that's not in the URL.

## Proposal

Add support for passing authentication tokens in the websocket subprotocol for Jupyter Server, to avoid tokens in URL parameters.

This is a scheme [devised by Kubernetes](https://github.com/kubernetes/kubernetes/commit/714f97d7baf4975ad3aa47735a868a81a984d1f0), where the subprotocols API allows specifying the Sec-Websocket-Protocol header, and we can put the token in there.

### Example implementation

(draft implementation of this at https://github.com/jupyter-server/jupyter_server/pull/1407)

To authenticate websockets via new mechanism, clients should:

```javascript
ws = new WebSocket(wss://..., ['v1.token.websocket.jupyter.org', `v1.token.websocket.jupyter.org.${token}`, ...])
```

which sets the header:

```
Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org, v1.token.websocket.jupyter.org.abc123
```

The response will have the header:

```
Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org
```

The reason for the double subprotocol is that if _any_ subprotocol is requested, the response _must_ include one of the requested subprotocols for the connection to be accepted by all browsers. Not all browser require this, but Chrome does. The `v1.token.websocket.jupyter.org` serves no purpose if there is already a subprotocol defined and required, and should be optional.

### Remaining questions

[...]

PR #121, opened 2024-04-30
JEP: Websocket token authentication with subprotocols

minrk:
JEP for #119

### DIFF
--- a/121-token-auth/token-auth.md
+++ b/121-token-auth/token-auth.md
@@ -1,0 +1,261 @@
+---
+title: Websocket token authentication with subprotocols
+authors: Min RK
+issue-number: 119
+pr-number: 121
+date-started: 2024-03-26
+---
+
+## Summary
+
+Jupyter servers shall accept auth tokens in the `Sec-WebSocket-Protocol` header,
+as is done for other API requests in the `Authorization` header,
+which is unavailable to websocket connections from browsers.
+The token shall be sent as:
+
+```
+Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org, v1.token.websocket.jupyter.org.{url-token}
+```
+
+where `{url-token}` is the url-encoded token (as produced by `encodeURIComponent` in javascript).
+
+This is fully backward-compatible.
+
+## Motivation
+
+Combining these facts:
+
+- Jupyter Servers (often) use API tokens for authentication
+- auth tokens are typically passed in the Authorization header
+- websockets cannot pass tokens in Authorization headers because the browser implementation [forbids it](https://github.com/whatwg/websockets/issues/16), though most implementations outside the browser do support it.
+  As a result, websocket requests must pass tokens in a URL parameter.
+- Passing tokens in a URL parameter is generally frowned upon, but not strictly insecure and indeed [explicitly recommended by Browser websocket implementers](https://github.com/whatwg/websockets/issues/16#issuecomment-347180825)
+
+motivates having a new mechanism by which to pass the auth token for websocket requests that's not in the URL.
+
+There is a scheme [devised by Kubernetes](https://github.com/kubernetes/kubernetes/commit/714f97d7baf4975ad3aa47735a868a81a984d1f0), where the subprotocols API allows specifying the Sec-Websocket-Protocol header, and we can put the token in there.
+
+## Guide-level explanation
+
+Many Jupyter server requests are authenticated with an API token.
+Currently, the API token may be specified in either the `Authorization` header or a `?token=` url parameter.
+Sending credentials in URL parameters is generally advised against, so `Authorization` is used in most API requests.
+However, the standard websocket API provided by browsers cannot set the `Authorization` header, so browser clients resort to sending the token in the url parameter in most cases.
+Websockets can, however, set the `Sec-Websocket-Protocol` header to a _list_ of values.
+
+This proposal adds a scheme for sending auth tokens in the `Sec-Websocket-Protocol` header.
+
+In general, this should be considered equivalent to sending tokens in the Authorization header which we do for all non-websocket API requests, but specifying the mechanism by which the token is transmitted when the for websocket requests when the Authorization header is unavailable.
+
+Adopting this scheme allows the removal of tokens from URLs, which is flagged by some security policies.
+
+### For clients
+
+Affected projects:
+
+- mainly `@jupyterlab/services`
+- any other client that connects to Jupyter Server websockets from browsers
+
+In the new authentication scheme, clients should request at least two subprotocols:
+
+- `v1.token.websocket.jupyter.org`
+- `v1.token.websocket.jupyter.org.${token}` where `${token}` is replaced by the token itself (url-encoded, as needed).
+
+For example:
+
+```javascript
+ws = new WebSocket(wss://..., ['v1.token.websocket.jupyter.org', `v1.token.websocket.jupyter.org.${encodeURIComponent(token)}`, ...])
+```
+
+which sets the header:
+
+```
+Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org, v1.token.websocket.jupyter.org.abc123
+```
+
+If the token is accepted, the response will have the header:
+
+```
+Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org
+```
+
+The reason for the double subprotocol is that if _any_ subprotocol is requested, the response _must_ include one of the requested subprotocols for the connection to be accepted by all browsers.
+Not all browser require this, but Chrome does. The `v1.token.websocket.jupyter.org` serves no purpose if there is already a subprotocol defined and required, and should be optional in that case.
+
+#### Backward compatibility
+
+This mechanism does not replace any other mechanisms, it is purely additional.
+A server that does not support the new scheme may reject a websocket connection with e.g. status 403, as if no token was provided.
+For clients to be backward compatible, they should first attempt the new scheme before falling back on an old scheme:
+
+1. try a connection with the new scheme,
+2. on failure
+   a. try again with an old scheme (e.g. `?token=...` in URL), or
+   b. raise error if requiring new scheme is desired
+
+There is already precedence for this retry pattern in JupyterLab for handling the kernel subprotocol.
+
+Clients that can send the token in the Authorization header may continue to do so without adopting the new scheme.
+The new scheme is specifically for clients that cannot set standard headers, i.e. browsers,
+it does not replace the use of the `Authorization` header where it is already available.
+
+### For servers
+
+Affected projects:
+
+- jupyter-server
+- jupyverse
+
+Before accepting a connection, a server should check the Sec-Websocket-Protocol header.
+If one of the protocols listed matches `v1.token.websocket.jupyter.org.{token}`, the token should be checked.
+The token should be url-decoded (e.g. `token = urllib.parse.unquote(token)`).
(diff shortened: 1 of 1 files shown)


121-token-auth/token-auth.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+If the token is accepted, the response will have the header:
+
+```
+Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org
+```
+
+The reason for the double subprotocol is that if _any_ subprotocol is requested, the response _must_ include one of the requested subprotocols for the connection to be accepted by all browsers.
+Not all browser require this, but Chrome does. The `v1.token.websocket.jupyter.org` serves no purpose if there is already a subprotocol defined and required, and should be optional in that case.
```
manics, line 83:
```suggestion
Not all browsers require this, but Chrome does. The `v1.token.websocket.jupyter.org` serves no purpose if there is already a subprotocol defined and required, and should be optional in that case.
```

121-token-auth/token-auth.md
```diff
@@ -0,0 +1,261 @@
...
+it does not replace the use of the `Authorization` header where it is already available.
+
+### For servers
+
+Affected projects:
+
+- jupyter-server
+- jupyverse
+
+Before accepting a connection, a server should check the Sec-Websocket-Protocol header.
```
manics, line 109:
```suggestion
Before accepting a connection, a server should check the `Sec-Websocket-Protocol` header.
```

121-token-auth/token-auth.md
```diff
@@ -0,0 +1,261 @@
...
+## Reference-level explanation
+
+### Clients
+
+Websocket clients SHALL transmit API tokens in the `Sec-Websocket-Protocol` header.
+
+- Token MUST be in the form `v1.token.websocket.jupyter.org.{token}`
+- Token MUST be url-encoded, e.g. via `encodeURIComponent(token)`
+- Subprotocol MUST include at least one OTHER subprotocol that is REQUIRED.
+  If no exiting subprotocol is REQUIRED, the subprotocol `v1.token.websocket.jupyter.org` MUST be included.
```
manics, line 134:
```suggestion
  If no existing subprotocol is REQUIRED, the subprotocol `v1.token.websocket.jupyter.org` MUST be included.
```

121-token-auth/token-auth.md
```diff
@@ -0,0 +1,261 @@
...
+
+- MAY ignore `token` in URL parameters
+
+#### Example implementation
+
+A draft implementation is submitted [to jupyter-server](https://github.com/jupyter-server/jupyter_server/pull/1407).
+
+## Rationale and alternatives
+
+### Following kubernetes example
```
manics, line 187:
```suggestion
### Following Kubernetes example
```

121-token-auth/token-auth.md
```diff
@@ -0,0 +1,261 @@
...
+
+This JEP follows a scheme [devised by Kubernetes](https://github.com/kubernetes/kubernetes/commit/714f97d7baf4975ad3aa47735a868a81a984d1f0).
+
+Our scheme differs from kubernetes only in that we do not serialize the token via base64 in the header.
```
manics, line 191:
```suggestion
Our scheme differs from Kubernetes only in that we do not serialize the token via base64 in the header.
```

121-token-auth/token-auth.md
```diff
@@ -0,0 +1,261 @@
...
+As a result, if a server doesn't implement subprotocol handling, clients must retry with an older scheme.
+
+This allows client and server security policies that may _reject_ auth tokens in URLs when the new scheme can be assumed.
+Specifying how/whether to do that is out of scope for this JEP.
+
+## Reference-level explanation
+
+### Clients
+
+Websocket clients SHALL transmit API tokens in the `Sec-Websocket-Protocol` header.
```
manics, line 129: Is the use of `SHALL` in some sentences and `MUST` in other sentences significant?

minrk: These SHALLs should be SHOULDs. I had [rfc2119](https://www.rfc-editor.org/rfc/rfc2119) in mind, where MUST is a requirement, while SHOULD is a recommendation. Sending tokens this way is not required because all existing mechanisms still work, but it is recommended where supported.

121-token-auth/token-auth.md
```diff
@@ -0,0 +1,261 @@
...
+Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org, v1.token.websocket.jupyter.org.{url_token}
+```
+
+where `{url_token}` is the url-encoded API token (note: in ~all cases in practice, `token == url_token`)
+
+If a subprotocol matching `v1.token.websocket.jupyter.org.{url_token}` is found:
+
+- `url_token` SHOULD be extracted and url-decoded (e.g. `token = unquote('{url_token}')`)
+- `token` SHOULD be handled identically to if it were sent via `Authorization: Bearer {token}`
+- If `token` is invalid or rejected, connection request MUST fail with status 403.
```
vidartf, line 167: Will it be clear to clients if the server didn't support the subprotocol or whether there was something wrong with the token? Ref other comment above, it seems like an overload of 403. If my token is expired, I don't want the client to fall back on tryin a less secure method.

vidartf (commented):
Positive for adding this support, as it's opt-in on both the server and client side. One technical question in diff comments.

121-token-auth/token-auth.md
```diff
@@ -0,0 +1,261 @@
...
+Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org
+```
+
+The reason for the double subprotocol is that if _any_ subprotocol is requested, the response _must_ include one of the requested subprotocols for the connection to be accepted by all browsers.
+Not all browsers require this, but Chrome does. The `v1.token.websocket.jupyter.org` serves no purpose if there is already a subprotocol defined and required, and should be optional in that case.
+
+#### Backward compatibility
+
+This mechanism does not replace any other mechanisms, it is purely additional.
+A server that does not support the new scheme may reject a websocket connection with e.g. status 403, as if no token was provided.
```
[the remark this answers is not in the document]
minrk: For reference, this bit _describes_ current behavior. We can propose _new_ behavior that doesn't support the new scheme (e.g. 401 on no creds), but if we are talking about backward compatibility, we have to handle what implementations do today without modification.

SylvainCorlay:
I guess that the retry pattern will just be a bit more complicated when dealing with the aligned kernel subprotocol?

minrk:
> I guess that the retry pattern will just be a bit more complicated when dealing with the aligned kernel subprotocol?

I'll need to check. In all cases, the first request should include the kernel subprotocol _and_ the token subprotocol, with the kernel subprotocol first indicating highest priority. There will be these possible cases to handle in terms of server support:

- supports both:
  - first request succeeds, subprotocol set to kernel subprotocol
- supports token but not kernel:
  - first request succeeds, subprotocol is set to token subprotocol - this indicates kernel subprotocol is _not_ supported
- supports kernel subprotocol but not token (this is currently released servers):
  - if cookie authentication available, succeeds and kernel subprotocol is set
  - if cookie authentication is not available, fails with 403. If this were a normal HTTP request, that would be easy to identify, but websockets obfuscate errors so it may be hard to distinguish from other failures
    - second request with kernel subprotocol and token in URL succeeds
- supports neither token nor kernel:
  - if cookie authentication available, fails with no supported subprotocol
    - (if auth failure is indistinguishable, which I believe it is) second request with token in url fails with no supported subprotocol
    - third request with token in url and no subprotocol succeeds
  - if cookie authentication is not available, fails with 403 (indistinguishable, I think)
    - same as above for the rest

Which means the first request reliably determines whether the token can be in the subprotocol. If the first request fails, the token shall be in the URL parameter. The second request then determines kernel subprotocol support (the first request in current jupyterlab), and in the event of a server supporting neither subprotocol, a third and final request is needed to make a successful connection.

So if token auth _is_ supported by the server, retries are no longer needed to check for support of the kernel subprotocol, which is a plus, I suppose. But in order to handle all possible cases, there could be up to 2 retries instead of the current 1, assuming I'm correct that clients can't meaningfully distinguish between a websocket that fails due to missing auth vs unsupported protocols. It could be simplified if the two failures turn out to be distinguishable in the client.

It may have been a good idea to define a subprotocol version string for the older wire format so that servers could explicitly declare that they only support the older wire format. But I'm not sure if that's worth discussing at this point, because that would reduce the number of cases where the retry is needed.

121-token-auth/token-auth.md
minrk: I realize the [comment below](https://github.com/jupyter/enhancement-proposals/pull/121#issuecomment-2176007180) is related to this: browsers don't report status codes, so there's no distinguishable difference for clients between 403, 404, 500, or any other reason a websocket request could fail (e.g. not a websocket endpoint at all). So status codes are not helpful for browsers (it's still the right thing to do for the server to log the right error code).

minrk:
I've run some tests, and browsers don't appear to record the http response (essentially, the WebSocket in browser seems to pretend that websockets are _not_ built on top of HTTP, so expose nothing about the HTTP requests/responses to client code). So clients need to treat _all_ connection failures as indistinguishable:

- no supported protocol
- auth error
- 404
- 500
- not a websocket endpoint at all
- lost connection

I'm not necessarily proposing we do this, but so folks can have an informed opinion, if the client _knows_ whether the token subprotocol is supported before attempting websocket requests, the conditions look like this:

- server supports token protocol:
  - first request fails: means auth error, no retry
  - first request succeeds:
    - selected subprotocol will be kernel if kernel subprotocol supported, token if kernel subprotcol not supported
- server doesn't support token protocol (token in url, same as now):
  - first request fails: auth error _or_ unsupported protocol
  - second request without kernel protocol:
     - succeeds if kernel subprotocol supported
     - fails if it was really an auth error or other problem

So it is simpler, especially since the current subprotocol retries can be eliminated if the token subprocol is known or assumed to be supported. But it adds the preflight specification to somehow communicate that token-authenticated websockets are supported, which we haven't decided on, and don't currently have a mechanism for.

vidartf:
@minrk is it possible to (ab)use the `reason` field in the close event? Or will that undermine some of the security considerations of websockets?

minrk:
> is it possible to (ab)use the reason field in the close event?

Yeah, I hadn't thought of that, but we could. I don't think it's abuse, I think it's actually what close code/reason are intended for. So instead of not accepting the connection in the first place, accept the connection and immediately close with a code (e.g. 4403 - 4000 + status code, since unregistered websocket codes should be in 4000-4999). This has an advantage in that it would actually give us a place for communicating the _reason_ for the close, which is helpful, and maybe what we should have been doing all along.

There are backward-compatibility downsides to the transition, at least:

- existing clients don't do this, so connection errors must still be handled by clients. But at least we should know that a connection error doesn't mean supporting token protocol + unauthorized token, if servers behave as intended
- from a Python perspective, it may be complicated to properly accept and immediately close connections without triggering unauthorized on_open/get/pre_get side effects, or accidentally accept unauthorized connections because currently `open` is assumed to be protected to only be called on authorized requests, and with such a change, `open` would end up being called (my idea here is to patch out `self.open = self.open_and_immediately_close`  and `self.get = WebSocketHandler.get`, but if overridden `close` assumes `open` has been called, errors are likely and hard to avoid.


It has a backward-compatibility downside because this is not what existing implementations do, so client still need to handle the current behavior.

It also has another backward-compatibility downside in that it moves the responsibility

You mean accept the connection and then immediately close with `reason`?

Issue #128, opened 2024-12-06
Pre-proposal: standardize object representations for ai and a protocol to retrieve them


mlucool:
# Summary

To deeply integrate AI into Jupyter, we should standardize both a method on objects to represent themselves and a messaging protocol for retrieving these representations. We propose using `_ai_repr_(**kwargs) -> str | dict` for objects to return representations. Additionally, we suggest creating a registry in kernels (e.g. IPython) for users to set representations for objects that do not define this method, along with a new message type for retrieving these representations.

# Motivation

Users should be able to include representations of instances of objects from their kernel as they interact with AI. This capability is what sets a productive Jupyter experience apart from other IDE-based approaches. For example, you should be able to use Jupyter AI and ask `Given @myvar, how do I add another row` or `What's the best way to do X with @myvar`?

While using something like `_repr_*` may have been sufficient, it can slow down display requests and does not allow passing information to hint about the shape of representations. For example, imagine a `Chart`. In a multimodal model, we may want to use a rendered image, but in a text-only model, we may want to pass only a description. Other model parameters or user preferences may also matter, such as the size of the context window or how verbose they want the representation to be.

Because of this, we suggest defining a new standard called `_ai_repr_(**kwargs) -> str | dict`. This method should return either a string or a MIME bundle. Additionally, since many libraries will not have this defined initially, there should be a registry where users can create a set of defaults and/or overrides, allowing them to use this feature without waiting for libraries to define it themselves.

Finally, the UI (e.g., jupyter-ai) needs a way to retrieve these representations for a given object. This is best done by introducing a new message type that can include the object and the kwargs. We expect this process to be slow at times (e.g., generating an image for a chart), so the control channel should not be used. Instead, a normal comms message can be used today, and as support for subshells improves, we can use that to avoid blocking while kernels are busy.

# Example
Continuing with the chart object example, we may want to add something like below. Typically this fictional chart it returns structured data for its JS display to render, but now we want an image for the context, which we expect to be slow to compute (e.g. a headless browser may need to be launched to do this):

```python
class JSBasedChart:
    ...

    def _ai_repr_(self, **kwargs):
        return {
            "text/plain": f"A chart titled {self.title} with series named {self.series_names}",
            "image/png": self.get_image()
        }
```

Other MIME types can also be used to enable the caller to represent the object in an optimized way for the model they are using (e.g., XML). For example, we could imagine Pandas DataFrame's defining this method:

```python
class DataFrame:
    ...
[...]

PR #129, opened 2025-02-07
AI Representation

mlucool:
In this JEP we propose:
1. A way for objects to represent themselves at runtime for AI
2. A registry for users to define representations for objects that do not have them
3. A new messaging protocol to query for this data

JEP for #128


mlucool:
CC @krassowski @ellisonbg @dlqqq @jtpio @SylvainCorlay @Zsailer @brichet

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,208 @@
...
+> Please generate a caption for @my_chart.
+
+It should be assumed that callers may choose to inject type infromation about these objects as well, making it
+even more powerful:
+> In @my_tbl, can you help me turn the columns labeled with dates into accumulated monthly columns
+
+In this case, `@my_tbl` would not only give the LLM data about its schema, but we'd know this is Pandas or Polars without
+a user having to specify this.
+
+It's possible that we'd want both an async and non-async version supported (or even just a sync version). If so, we can default one to the other:
```
dlqqq, line 63: It may be better to have the async version live in a different method, e.g. `_async_ai_repr_()`.

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,208 @@
...
+        "metadata": {
+            "<mime-type>": {              # Metadata associated with specific MIME types
+                "<key>": "<value>"
+            }
+        },
+        "transient": {
+            "display_id": "<str>"        # Optional runtime-specific metadata
+        }
+    }
+}
```
dlqqq, line 132:
````suggestion
}
```
````

File: 129-ai-representation/ai-representation.md [hunk 3]
```diff
@@ -0,0 +1,208 @@
...
+    "content": {
+        "object_name": "<str>",        # The name of the object whose representation is requested
+        "kwargs": {                     # Optional context-specific parameters for representation customization
+            "<key>": "<value>"
+        }
+    }
+}
+```
+- **`object_name`**: (Required) The name of the variable or object in the kernel namespace for which the representation is being requested.
+- **`kwargs`**: (Optional) Key-value pairs allowing customization of the representation (e.g., toggling multimodal outputs, adjusting verbosity).
```
dlqqq, line 111: It seems like the `kwargs` structure needs to be documented somewhere. Would this be standardized by the kernel implementation when handling `ai_repr_request`?

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,208 @@
...
+- Develop AI-specific registries for popular libraries (e.g., pandas, matplotlib) or make PRs to add this natively.
+
+This proposal lays the groundwork for seamless AI integration in Jupyter, empowering users to derive greater insights and efficiencies from their notebooks.
+
+# Unresolved questions
+
+- Should we support both or one of async/sync `_repr_ai_`
+- What are good reccomended kwargs to pass
+- How should this related to `repr_*`, if at all.
+- What is the right default for objects without reprs/formatters defined? `str(obj)`, `None`, or `_repr_`?
```
dlqqq, line 206: A registry can define AI representations for objects that lack a `_ai_repr_()`. It seems natural to also a registry to define a "fallback" AI representation, i.e. the method to be used when neither a `_ai_repr_()` method or a registry entry exist for an object.

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,208 @@
...
+- Should thread-saftey be required so that this can be called via a comm
+- Can `ai_repr_request` be canceled?
```
dlqqq, line 208:
```suggestion
- Should we support both or one of async/sync `_ai_repr_()`?
- What are good recommended `kwargs` to pass in `ai_repr_request`?
- How should this relate to `repr_*`, if at all?
- What is the right default for objects without reprs/formatters defined? `str(obj)`, `None`, or `_repr_`?
- Should thread-safety be required so that this can be called via a comm?
- Can `ai_repr_request` be canceled?
```

dlqqq (commented):
@mlucool @govinda18 Thank you for opening this JEP! Really excited to see this moving forward. 💪 

Left some feedback & typo fixes below.

Carreau:
I think there was a bug in github (or my laptop), I submitted my review yesterday, but there was no internet, and it resubmitted today when I reopened the tab and internet was back. Sorry if there are crossed wires.

129-ai-representation/ai-representation.md
Carreau: my understanding is that they should be some consensus, but some free parameters that could depends on which repr understand the need of which models.

129-ai-representation/ai-representation.md
Carreau: My guess is the registry is already the fallback, because the object does not have a `_ai_repr_`, Unless you see the registry as an override of `_ai_repr_` ?

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,208 @@
...
+    "type": "ai_repr_request",
+    "content": {
+        "object_name": "<str>",        # The name of the object whose representation is requested
+        "kwargs": {                     # Optional context-specific parameters for representation customization
+            "<key>": "<value>"
+        }
+    }
+}
+```
+- **`object_name`**: (Required) The name of the variable or object in the kernel namespace for which the representation is being requested.
```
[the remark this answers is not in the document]
mlucool: I think that's ok too

129-ai-representation/ai-representation.md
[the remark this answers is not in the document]
mlucool: > There is also the slight technical question that _ai_repr_(**kwargs) -> Dict[str, T], requires T to be serialisable, no really be Any.

Agreed

129-ai-representation/ai-representation.md
mlucool: My intent was not to be prescriptive here and let that evolve naturally. That is, given `jupyter-ai` popularity, I expect it to implicitly set some kwargs as a standard. No kwarg should be required is maybe a better statement here, but plugins are free to document things they support.

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+This allows flexibility:
+- For text-only models, the `text/plain` key provides a concise description.
+- For multimodal models, `image/png` can render the chart visually.
+
+For specific extensions, they can define new mime-types that give the required data.
+
+In practice, the kwargs should guides the `_ai_repr_` determine what to include in the mime bundle. For a text-only
+model, we should not waste time generating the image.
+
+Now I can talk with my an mutli-modal LLM. For example, I may say (assuming my plugin substitutes @ for the value form `_ai_repr_`):
```
krassowski, line 53:
```suggestion
Now I can talk with my mutli-modal LLM. For example, I may say (assuming my plugin substitutes @ for the value form `_ai_repr_`):
```

krassowski (commented):
This was discussed in the jupyter-ai call today.

We just wanted to ping CC @jupyter/software-steering-council on this :)

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+title: AI Representation Protocol for Jupyter
+authors: Marc Udoff, Govinda Totla
+issue-number: 128
+pr-number: 129
+date-started: 2025-02-07
+---
+
+### Summary
+
+This proposal introduces a standardized method, `_ai_repr_(self, **kwargs) -> Dict[str, Any]`, for objects in Jupyter environments to provide context-sensitive representations optimized for AI interactions. The protocol allows flexibility for multimodal or text-only representations and supports user-defined registries for objects lacking native implementations. Additionally, we propose a new messaging protocol for retrieving these representations, facilitating seamless integration into Jupyter AI tools.
```
minrk, line 11: Absolutely minor detail, but is there a reason this inverts the standardized `_repr_ai_` naming scheme to `_ai_repr_`? If not, then I'd suggest following the prefix precedent  (using a prefix improves discoverability via tab completion, etc.).

minrk (commented):
This seems generally sensible

My main design question is at the protocol level, why does this require a distinct message handler, as opposed to defining a schema for one (or more) custom `x-jupyter/ai-repr` mimetype via the existing display protocol, e.g. an `inspect_request` which already returns a mimebundle representing an object.

I don't really understand why the existing mimebundle messages have "performance concerns and their inability to adapt" while this doesn't, when it appears to do the same thing (returns a mimebundle representing an object), and has the same design (register reprs methods both via method name and explicit dispatch).

krassowski:
> via the existing display protocol, e.g. an `inspect_request` which already returns a mimebundle representing an object.

`inspect_request` takes `code`, `cursor_pos` (position in given code) and `detail_level` (0 or 1). I proposed re-using this before, some of the counter arguments were:
- it does not allow to pass extra keyword arguments
- `detail_level` is poorly defined, and LLMs will often need a hard cutoff on number of tokens they want in representation as this it is proportional to cost/performance
- sending `code` and `cursor_pos` is a design deriving from "user has a cursor in a cell", not from "user asks about a variable in the chat". `cursor_pos` could be always set to 0 and `code` could represent the variable so it is not a strong argument

Now, another message already in the protocol that could be considered is `richInspectVariables` which is an addition to the Debugger Adopter Protocol. It is defined as:

<details>

```python
{
    'type' : 'request',
    'command' : 'richInspectVariables',
    'arguments' : {
        'variableName' : str,
        # The frameId is used when the debugger hit a breakpoint only.
        'frameId' : int
    }
}
```

and returns:

```python
{
    'type' : 'response',
    'success' : bool,
    'body' : {
        # Dictionary of rich representations of the variable
        'data' : dict,
        'metadata' : dict
    }
}
```

</details>

The advantage of using frame IDs is that local variables can be distinguished from global variables. Now, the problem is that debugger is opt-in, has side-effects and performance overhead so it is not necessarily a go-to solution. The basic DAP uses all three: `name`, `variablesReference`, and `frameId` to determine the variable being requested.

I wonder if when debugger is active, should we have a way to pass `frameId` and maybe `variablesReference` to allow user to ask about contents of a local vs non-local or global variable?

krassowski:
A broader edge-case is definition disambiguation:
- Julia and R use multiple dispatch natively
- Python has definition overloads with `@overload` decorator
- the same name in TypeScript can refer to both a namespace and a class

While it is fine to return description of all cases, it can grow long. In these cases to get a definition of a function unambiguously we would need to know both the function and the argument(s); This is where `inspect_request`'s way of doing things shines. Of course if would require the user to provide more context than just the function name.

This JEP does not say how user would specify `@var`. My assumption is that UI could use `complete_request` request to get a list of variables as user starts typing. If that's the case one solution to "which definition is it referring to" could be augmenting `complete_reply` to return a unique `variablesReference` for each completion. Imagine user asks:

```
What are the arguments for @predict
```

In R if they loaded a few statistical packages, they could be presented with a list dozens of different variants; a simple case (for brevity):

```R
# S3 method for glm
predict(object, newdata = NULL,
            type = c("link", "response", "terms"),
            se.fit = FALSE, dispersion = NULL, terms = NULL,
            na.action = na.pass, …)

# S3 method for lm
predict(object, newdata, se.fit = FALSE, scale = NULL, df = Inf,
        interval = c("none", "confidence", "prediction"),
        level = 0.95, type = c("response", "terms"),
        terms = NULL, na.action = na.pass,
        pred.var = res.var/weights, weights = 1, …)
```

Now, we may want to present these as separate items in `complete_reply` for a few reasons:
- each has a different signature and documentation, concatenating all of these would be too long to display in completion panel
- each signature defines a different completion snippet (think template for filling in arguments, see [Snippet Syntax](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax) for LSP way of doing it)

Then if a user chooses one or another, a different snippet would show up. These separate items could be annotated with a variable reference, so when it is used for AI request proposed in this JEP the user's choice would be remembered.

I think defining a variable reference in `complete_reply` and as an argument for `ai_repr_request` does not need to be a part of this JEP, but I think it highlights the difference from `inspect_request` which could not achieve the same easily, unless it's signature is changed in ways which I would suspect would be called breaking/requiring a major release.

ivanov:
AI is such a nebulous term that whatever description or formulation this representation would hold could both be useful to humans, and also useless to past, present, and future applications that would still nevertheless be reasonably called AI. So I came here to chime in that this can already be done as a set of conventions around mimebundle keys, and hence not require a protocol revision.

Fortunately, @minrk is already doing the same
> I don't really understand why the existing mimebundle messages have "performance concerns and their inability to adapt" while this doesn't, when it appears to do the same thing (returns a mimebundle representing an object), and has the same design (register reprs methods both via method name and explicit dispatch).

We need to be very deliberate and conservative with protocol revision, I think this would grow the surface area and complexity of interactions for the jupyter protocol. For example, if you end up with objects that declare "ai repr" but not other rich repr, do you end up with users who aren't using "ai" having to inspect objects twice in separate ways, hoping to get something better than `<Foo object at 0x7f600ecf8990>` ?

129-ai-representation/ai-representation.md
mlucool: I believe this came from chatting with @Carreau and we wanted to avoid it ever being automatically called during display. (FWIW aside from this, if its not supposed to go into the general mime bundle, it feels strange to use the same pattern).

@Carreau do you recall the exact details?

mlucool:
> AI is such a nebulous term that whatever description or formulation this representation would hold could both be useful to humans, and also useless to past, present, and future applications that would still nevertheless be reasonably called AI.

I agree it is nebulous, do you have a better term we should use? I started with calling it `repr_llm`, but `llm` seemed too specific. 


> I don't really understand why the existing mimebundle messages have "performance concerns and their inability to adapt" while this doesn't, when it appears to do the same thing (returns a mimebundle representing an object), and has the same design (register reprs methods both via method name and explicit dispatch).

This is a important question and I believe there are a number of reasons
1.  If this is included in the mimebundle, then all displays of an object must also spend time computing an AI centric display of the data. This information is rarely needed to render an object in a notebook, but is useful for representing an object in an context window. This makes things slower to render in notebook and sends potentially a large amount of extra data.
2. I think an async version is useful. Imagine you want to render a JS chart in notebook (e.g. plotly). In the mime bundle you'll want to pass datapoints etc., and the front end has the chart library render. Now if I want to use this chart with an LLM, I'd want an image. This is a costly process to compute and likely happens outside the kernel (e.g. send it to playwright).
3. I believe you still need a new protocol for the reasons @krassowski outlines above (even if we go with this is just a mimebundle).

That all being said, 1/2 are a bit unrelated to the protocol change (we still need a way to get this data from UIs). If we came up with other solutions here (e.g. an async mimebundle, where there was a clean way to pass args to), I don't think it needs to be "ai".

mlucool:
> Would a custom mime-type in existing messages be satisfactory if it could be opt-in via the request, and not computed by default? Then the same request that asks for the "genai repi" could explicitly exclude all the other reprs, too. That appears to address all of the performance/efficiency questions, IIUC.

Not a tangent at all. It's a great question. Here's a first pass at what I think would have to change:
1. We still need a new message to reach into a kernel and call this method for the reasons [described above](https://github.com/jupyter/enhancement-proposals/pull/129#issuecomment-2687413017).
2. It's not just about opting in/out but also letting kwargs flow through to the bundle. We'd want to really start using the optional kwargs, which seems fine, but is a departure to how this is called. I'm not sure the impact on existing code
4. The merits of an async version should still be discussed. Maybe I am over indexing on Chart->image, but I suspect it'll be really common. It's also possible that subshells is a cleaner way to allow for compatibility.

minrk:
Thanks for answering, I think the kwargs was the big missing piece I didn't understand before, which is more different than my initial interpretation.

mlucool:
>  I think the kwargs was the big missing piece I didn't understand before, which is more different than my initial interpretation.

Great! @minrk is there anything else I can answer to help clarify the JEP?

minrk:
@mlucool thanks, I think my questions have been addressed.

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
+---
+title: AI Representation Protocol for Jupyter
+authors: Marc Udoff, Govinda Totla
+issue-number: 128
+pr-number: 129
+date-started: 2025-02-07
+---
+
+### Summary
```
krassowski, line 9:
```suggestion
# AI Representation Protocol for Jupyter

## Summary
```

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+
+This proposal introduces a standardized method, `_ai_repr_(self, **kwargs) -> Dict[str, Any]`, for objects in Jupyter environments to provide context-sensitive representations optimized for AI interactions. The protocol allows flexibility for multimodal or text-only representations and supports user-defined registries for objects lacking native implementations. Additionally, we propose a new messaging protocol for retrieving these representations, facilitating seamless integration into Jupyter AI tools.
+
+### Motivation
```
krassowski, line 13:
```suggestion
## Motivation
```

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+
+Expected outcomes include enhanced AI-driven workflows, better support for multimodal interactions, and improved user experience in Jupyter-based AI applications.
+
+While this JEP focuses on Jupyter, it's understood that others may choose to reuse this same function in any python
+process. While that would be a good outcome for the community, it is not the goal of this repr to set standards in
+Python in general.
+
+### Guide-Level Explanation
+
+#### Introducing the `_ai_repr_` Protocol
```
krassowski, line 29:
```suggestion
## Guide-Level Explanation

### Introducing the `_ai_repr_` Protocol
```

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+a user having to specify this.
+
+It's possible that we'd want both an async and non-async version supported (or even just a sync version). If so, we can default one to the other:
+```
+class CustomChart:
+    async def _ai_repr_(self, **kwargs):
+        return _ai_repr_(**kwargs)
+```
+
+#### Using Registries for Unsupported Objects
```
krassowski, line 70:
```suggestion
### Using Registries for Unsupported Objects
```

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+
+This allows users to make useful representations of all objects they care about now, and extend the mime type of
+existing objects without needing to monkey patch anything. This layering is imporant for empowering users quickly in this
+space, without forcing any project to adopt nor forcing any user to upgrade a library.
+
+The registry should be free of any dependency to minimize conflicts.
+
+Both the registry and repr on objects are intended to be fully backwards compatible.
+
+#### Messaging Protocol
```
krassowski, line 95:
```suggestion
### Messaging Protocol
```

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+}
+```
+- **`data`**: A dictionary of MIME-typed representations for the object.
+- **`metadata`**: A dictionary of metadata associated with the data, including global metadata and MIME-type-specific sub-dictionaries.
+- **`transient`**: A dictionary of transient metadata, not intended to be persisted, such as a `display_id` for session-specific tracking.
+
+
+##### Example
```
krassowski, line 139:
```suggestion
#### Example
```

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+```
+
+This allows for front end plugins enable the variable reference features (`@myvar`) noted above. Because it can pass kwargs,
+it has fine-grained control over passsing relevant information to reprs.
+
+### Reference-Level Explanation
+
+#### Design Details
```
krassowski, line 156:
```suggestion
## Reference-Level Explanation

### Design Details
```

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+- **Extend `_repr_*` methods**: Lacks flexibility for passing model specific features (e.g., multimodal) and context-specific needs (e.g., using more/less of a context window).
+- **Functional API only**: A single top-level function could work but limits extensibility and modularity.
+- **Ignore standardization**: Risks fragmented, incompatible implementations.
+
+### Prior Art
+
+- **IPython’s `_repr_*` methods**: Provides the foundation but lacks AI-specific extensions.
+- **Multimodal frameworks**: Other ecosystems (e.g., Hugging Face) emphasize context-aware representations.
+
+### Future Possibilities
```
krassowski, line 196:
```suggestion
### Interactions with Existing Features

- Builds on Jupyter’s existing `_repr_*` ecosystem but focuses on AI-specific needs.
- Compatible with existing kernel and frontend architectures.

## Rationale and Alternatives

### Rationale

- Standardization ensures consistent, predictable behavior across libraries.
- Flexible MIME bundles cater to diverse AI model requirements.
- Registries accelerate adoption without waiting for upstream changes.

### Alternatives

- **Extend `_repr_*` methods**: Lacks flexibility for passing model specific features (e.g., multimodal) and context-specific needs (e.g., using more/less of a context window).
- **Functional API only**: A single top-level function could work but limits extensibility and modularity.
- **Ignore standardization**: Risks fragmented, incompatible implementations.

## Prior Art

- **IPython’s `_repr_*` methods**: Provides the foundation but lacks AI-specific extensions.
- **Multimodal frameworks**: Other ecosystems (e.g., Hugging Face) emphasize context-aware representations.

## Future Possibilities
```

129-ai-representation/ai-representation.md
```diff
@@ -0,0 +1,209 @@
...
+
+- Develop AI-specific registries for popular libraries (e.g., pandas, matplotlib) or make PRs to add this natively.
+
+This proposal lays the groundwork for seamless AI integration in Jupyter, empowering users to derive greater insights and efficiencies from their notebooks.
+
+# Unresolved questions
```
krassowski, line 202:
```suggestion
## Unresolved questions
```

krassowski (commented):
This is just to align headings with the fixes in https://github.com/jupyter/enhancement-proposals/pull/130.

Also, can you merge with `master` branch to pull the fixes for build?

westurner:
Would a JSON-LD 'ified nbformat and` _repr_jsonld_()` solve for this too?

- https://github.com/jupyter/nbformat/issues/44

westurner:
MCP: Model Context Protocol; https://modelcontextprotocol.io/introduction :

> MCP is an open protocol that standardizes how applications provide context to LLMs. Think of MCP like a USB-C port for AI applications. Just as USB-C provides a standardized way to connect your devices to various peripherals and accessories, MCP provides a standardized way to connect AI models to different data sources and tools.

https://github.com/modelcontextprotocol

Are there other similar standards for what the MCP specification solves for?

mlucool:
I believe MCP to a bit tangent to this. MCP is more for agents tools (i.e. something that can do work) and this is meant to be a way to call for specific information for a specific variable.

JohanMabille:
Having a dedicated message makes it simpler for kernels to implement it and is backward compatible, at the cost of potentially duplicating requests  (inspect and ai_repr).

Another solution could be to add optional arguments to the inspect_requests, that could be used to specify the user wants an AI  representation, and would be unused in the case of a classical inspect_request. Similarly the `code` and `pos` arguments of the inspect_request would be unused in the cas of an ai_repr. This would avoid the scenario where the front end has to ask twice the same representation of an object depending on the context, and would add flexibility to the inspect_request (the front end can specify which representation or combination of representation he wants). This is not the solution I prefer because it looks like we are using the same messages for two different (although similar) things and the message will have optional argument depending on the use case.

In both cases kernels can advertise they support this new feature via the [Optional Feature](https://github.com/jupyter/enhancement-proposals/blob/master/92-jupyter-optional-features/jupyter-optional-features.md), making it possible to have a backward compatible protocol extension.

westurner:
Use cases; an MCP server for Jupyter:

- Src: https://github.com/block/mcp-jupyter
- docs: https://block.github.io/mcp-jupyter/
- PyPI: https://pypi.org/project/mcp-jupyter/

> Jupyter MCP Server allows you to use tools like Goose or Cursor to pair
with you in a JupyterLab notebook where the state of your variables, etc is
preserved by the Jupterlab Kernel. The fact that state is preserved is the
key to this because it allows to to pair with the Agent in a notebook,
where for example if a package is not installed it will see the error and
install it for you. You as the user can then do some data exploration and
then hand off to the agent at any time to pick up where you left off.
>
> This works with any client that supports MCP but will focus on using
Goose for the examples.

rgbkrk:
Could we have `_ai_repr_` go to a specialized `text/llm+plain` so that it's differentiable from classic `text/plain` and can technically be more verbose than a regular plain repr?

westurner:
I wrote a quick tool to transform ai chat exports from .md/.json to .ipynb
a few weeks ago; "transform_md" that still haven't factored extra
functionality out of;

ipynb supports multiple mimetype outputs, if they have different types; but
not multiple `text/html` outputs per input?

Where to display the model's thinking as output?

- input
- output_thinking
- output_response

.

- I: input
- O1: output_thinking
- O1: output_response

.

- I1: input
- O1: output_thinking
- I2:
- O2: output_response

It's clunky to try and copy/paste just an output_response if it's
string-concatenated with output_thinking.

GitHub Copilot JSON is a more complete trace than most other chat AI export
formats?

Is there already a standardized JSON-LD/YAML-LD "Open Trace" format that
covers?

That can support *multiple*  _ai_repr_ outputs per input? With which MIME
types?

And then, just like notebooks in Markdown, copying binary files inlines
into an .ipynb file is clunky. IIRC the markdown nbformat issue(s) already
discuss how moving from .ipynb to .zip basically implies that you're
creating or adopting a new packaging format (when all they want is support
for pyproject.toml, requirements.txt, and also environment.yml, and PEP723
inline script metadata, repo2docker/repo2jupyterlite and container2wasm)


- "feat(cli): Add /export command to export chat history to markdown and
jsonl"
https://github.com/google-gemini/gemini-cli/pull/5342#pullrequestreview-3912374169
- "How to export the chat history of GitHub Copilot Chat?"
https://github.com/orgs/community/discussions/57190

- How many outputs per input?
- Which MIME types for the _repr_ai_ outputs?
- How/when to merge all of the JSON outputs into one document?
- How to merge all of the JSON-LD outputs?
- How to merge all of the _repr_json_ and _ai_repr_ outputs?
How to merge all of the _repr_json_ and _ai_repr_ outputs, with notebook
level metadata and cell level metadata; into one JSON document with a
JSON-LD @context?

.

- The other methods have the prefix `_repr_` as a convention, so `_repr_ai`
would be better

rgbkrk:
I've been stewing on this a bit and I don't really like the addition of a protocol to get what is effectively a summary and call it ai. It's clearly intended to be used by AI but is more like a variable inspection request where we pull summary data. There's several areas lurking at once in this JEP that I think are easy to push on without having to make a kernel protocol as we can use existing machinery or create more direct building blocks on top of existing ones.

westurner:
How to indicate the schema version of what is returned by _repr_json_ or
_repr_jsonld_ or _repr_yamlld12_ or _repr_ai_?

Or just gather each from all output cells into a document after each cell
runs

How to publish linked data for linked research, from: document level
metadata including kernelspecs, cell level metadata, cell output data?

- [x] schema.org/CreativeWork
- [x] schema.org/DigitalDocument
- [x] schema.org/ScholarlyArticle
- [ ]  schema.org/Notebook ,
       schema.org/DigitalNotebook
- [ ] schema.org/JupyterNotebook
                or just Notebook

Justify adding _repr_ai_ or linked data support as described

PR #130, opened 2025-03-05
Fix missing JEPs, move `kernel-subshells` to follow convention

krassowski:
Also fixes syntax error in `Kernel Handshaking pattern` link on the website and moves Notebook v7 to "Implemented' in ToC (it was marked as Implemented in README in my previous https://github.com/jupyter/enhancement-proposals/pull/124).

| Before | After |
|--------|--------|
| image | image |

| Before | After |
|--------|--------|
| image | image |

Changes in the table shading and search bar are due to newer version of sphinx/sphinx-book-theme.


krassowski:
A few of the JEPs did not follow the template and need to be updated to display properly in the ToC:

krassowski:
I don't know if SSC at some point decided not to add numerical prefixes any more and not to update the website, but given that almost none of the JEPs accepted in last two years are discoverable via https://jupyter.org/enhancement-proposals/ which is the public facing index of JEPs, I think this was not a good idea.

krassowski:
Probably a number of JEPs which I added to ToC/READMEas "Accepted" are already implemented, but I think this PR is already large and I do not want to hold it for discussions on whether something is implemented or not (like subshells are partially implemented).

krassowski:
CC @jupyter/software-steering-council please review.

minrk (approved):

krassowski:
Thanks for the approval! I do not have merge rights here, so someone from SSC or EC will need to do that.

Zsailer:
Thanks @krassowski! And thank you @minrk for the review. Merging.

PR #138, opened 2026-01-30
Draft: JEP: Specify the Markup Language and flavor

rgaiacs:
This is related to the pre-proposal discussed in https://github.com/jupyter/enhancement-proposals/issues/98.

This is a counter proposal to https://github.com/jupyter/enhancement-proposals/pull/99.

As this is a draft, feedback is welcome!


xx-markup-language/markup-language.md
```diff
@@ -0,0 +1,204 @@
...
+```json
+{
+    "cell_type": "markup",
+    "source": "[some multi-line markup]",
+    "outputs": {
+        "text/plain": "[multiline plain text output]",
+        "text/html": "[multiline html text output]",
+    }
+}
+```
```
krassowski: I like this proposal in general. However, introducing a new cell type is a big ask and will result in new notebooks failing to display at all in many renderers.

Question: could this be achieved with keeping the existing cell types?
- could we extend the `markdown` cell type (this would allow to render most of Myst/quatro markdown in existing renderers): is there a markup language which is not markdown that we would want to support?
  - I imagine HTML is technically not markdown, but markdown can contain HTML so its moot.
  - I imagine someone could define markup for LaTeX.
- could we extend `code` cell type: is there a markup language which is not really some sort of `code`? HTML is code, so markdown and LaTeX could be called code if we stretch the definition a bit. Since this proposal suggests adding `outputs`, this is what `code` cells already have
- could we extend `raw` cell type?

If there are strong reasons not to use any of these three then I think it is fair to introduce `markup`, but this proposal should provide a strong rationale for it.

rgaiacs: Thanks for the feedback.

> could we extend the `markdown` cell type

It would work for me for backward compatibility.

> is there a markup language which is not markdown that we would want to support?

Some clarification to avoid confusion with the term "support". This pull request / proposal does not intend to increase the work for Jupyter team maintains. The goal of this pull request / proposal is to enable a more diverse work to be done in the Jupyter ecosystem. Support for markup languages should be treated the same way as support for [kernels](https://docs.jupyter.org/en/stable/projects/kernels.html). The community would be responsible for maintain support for markup languages other than Markdown using extensions.

"Full" LaTeX support is valuable for researchers / academics who are writing their manuscript using LaTeX.

My impression is that the use of [reStructuredText](https://en.wikipedia.org/wiki/ReStructuredText) is in decline because of [MyST](https://mystmd.org/) but would be interesting to see reStructuredText support.

> could we extend code cell type

This also makes sense to me. Unfortunately, this will break backward compatibility because a Jupyter Notebook would have a `code` cell with Python code and another `code` cell with Markdown. Existing clients would send the `code` cell with Markdown to the Python kernel and the user would receive back a error message. @krassowski what is your option about this?

But using `code` cell makes sense given that inline expression evaluation is supported by [MyST](https://mystmd.org/guide/notebooks-with-markdown#myst-inline-expressions) and [Quarto](https://quarto.org/docs/computations/inline-code.html).

> could we extend raw cell type

I think the other two options are better.

rgaiacs:
Thanks for the positive feedback. I amended the enhancement proposal to avoid a new cell type and preserve backward compatibility.

xx-markup-language/markup-language.md
agoose77: See also https://github.com/jupyter/enhancement-proposals/blob/master/97-add-schema/add-schema-to-notebook-format.md

xx-markup-language/markup-language.md
```diff
@@ -0,0 +1,205 @@
...
+        },
+        "language_info": {
+            # if language_info is defined, its name field is required.
+            "name": "the programming language of the kernel",
+            "version": "the version of the language",
+            "codemirror_mode": "The name of the codemirror mode to use [optional]",
+        },
+        "markdown_info": {
+            # if markdown_info is defined, its name field is required.
+            "name": "the markup language of the kernel",
```
rgaiacs, line 60: @agoose77 suggested to use media type here. My understand is that we would have to use [unregistered tree](https://en.wikipedia.org/wiki/Media_type#Unregistered_tree), for example, `text/x.myst` or `text/x.pandoc; footnotes=true`.

I don't like the suggestion because the `name` was think to be similar to the `name` of the `kernel_info`, i.e. a human readable version.

Issue #75, opened 2021-09-21
pre-proposal: CurveZMQ


minrk:
zeromq has a transport-level encryption and authentication protocol called [CurveZMQ][].

[CurveZMQ]: https://rfc.zeromq.org/spec/26/

I've just landed support for CurveZMQ in [IPython Parallel][ipp-curve], and I think it's worth talking about in Jupyter.

[ipp-curve]: https://github.com/ipython/ipyparallel/pull/553

The gist of the most basic implementation:

- **server** sockets set a public and private key, issued e.g. via `zmq.curve_keypair()` from pyzmq. (the sockets that bind make the most sense, but is not strictly a requirement), and `CURVE_SERVER=1` to enable auth.
- **client** sockets use the server's public key as a **server key** - anyone with this key can connect to the server and send/receive messages
- client's own private/public key can be any key pair, and are used exclusively for encryption; they have no role in authentication

In Jupyter, we already have a key distribution mechanism, which is the HMAC message-signing key in connection files. We can use the same key distribution for Curve keys. In the context of Jupyter, it's a little weird, because it's usually the client (KernelManager) that *sets* the credentials, which means the client issues the kernel's private key as well, and needs to pass the *private* key to the kernel. This being the case, the absolute simplest version is to use the same private/public keypair for both ends.

### Sketch:

- KernelManager issues private, public keypair with `zmq.curve_keypair()`
- store private_key and public_key in connection file
- kernel sockets set CURVE_PRIVATEKEY, CURVE_PUBLICKEY, CURVE_SERVER=1
- client sockets set the same CURVE_PRIVATEKEY, CURVE_PUBLICKEY, and use the public key in CURVE_SERVERKEY (alternative: client sockets issue new private/public keys, but if the private key is already in the connection file, I see no benefit to issuing more keypairs)

Here's an example of an authenticated socket pair in pyzmq:

<details>

<summary>curve socket example</summary>

```python
import asyncio
import zmq
import zmq.asyncio as zaio

async def main():
    public, private = zmq.curve_keypair()
    ctx = zaio.Context()
    server = ctx.socket(zmq.ROUTER)
    # server socket is a 'curve server'
    server.CURVE_SECRETKEY = private
    server.CURVE_PUBLICKEY = public
    server.CURVE_SERVER = True

    url = "tcp://127.0.0.1:5555"
    server.bind(url)

    no_auth_client = ctx.socket(zmq.DEALER)

[...]

PR #145, opened 2026-06-10
Transport encryption for ZMQ communication

krassowski:
1. [x] **Create a GitHub issue**: #75 
2. [x] **Identify a Shepherd**: @JohanMabille 
3. [x] **Decide if it's a JEP** according to the criteria listed above. The Shepherd decides if the JEP criteria have been met.
4. [x] **Used the new JEP markdown template** to help structure your thoughts. You can find it in `jupyter-enhancement-proposal-guidelines/TEMPLATE.md` in this repository.

Q: Is there a way to view the preview from the `deploy-book` action?

Proof of concept implementation:
- https://github.com/jupyter/jupyter_client/pull/1110
- https://github.com/jupyter/jupyter_client/pull/1124
- https://github.com/ipython/ipykernel/pull/1515
- https://github.com/jupyter-server/jupyter_server/pull/1638
- https://github.com/jupyter-server/gateway_provisioners/pull/147

JEP alignment PRs:
- https://github.com/jupyter/jupyter_client/pull/1126
- https://github.com/jupyter/jupyter_client/pull/1127
- https://github.com/ipython/ipykernel/pull/1528
- https://github.com/jupyter-server/jupyter_server/pull/1660


krassowski:
CC @minrk and @Carreau for initial review before I undraft it.

@JohanMabille would you like to be a shepherd of this JEP?

JohanMabille:
> @JohanMabille would you like to be a shepherd of this JEP?

Sure

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,226 @@
...
+
+# Transport encryption for ZMQ communication
+
+## Summary
+
+We propose to support encrypting the traffic exchanged between the kernels and client, adding a dedicated stanza in `kernelspec` to enable kernels to communicate support for encryption, adding a new set of fields for key exhange in the connection file, and using CurveZMQ protocol as a default implementation of such an encryption.
+
+## Motivation
+
+With the default TCP transport non-authenticated clients with access to ports on the same machine are currently able to monitor IOPub, accessing any secrets exchanged during kernel runtime (code, outputs, etc). It can be only prevented by using ICP transport and file permissions.
```
minrk:
```suggestion
With the default TCP transport non-authenticated clients with access to ports on the same machine are currently able to monitor IOPub, accessing any secrets exchanged during kernel runtime (code, outputs, etc). It can be only prevented by using IPC transport and file permissions.
```

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,226 @@
...
+
+There are two new diagnostics a user may notice. When a kernel runs over TCP _without_ encryption, `ipykernel` now logs a warning:
+
+> Kernel is running over TCP without encryption. All traffic, including any secrets in cell outputs, is readable by other processes on this machine. Use the IPC transport or start the kernel through a manager that provisions CurveZMQ keys.
+
+And requesting `transport_encryption` of `required` for a kernel that does not advertise support fails at startup:
+
+> RuntimeError: transport_encryption='required' but kernelspec does not declare `metadata.supported_encryption='curve'`.
+
+We recommend other kernels that decide to implement encryption to emit similar warning and error messages.
```
minrk:
```suggestion
We recommend other kernels and clients that decide to implement encryption to emit similar warning and error messages.
```

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,226 @@
...
+
+  ```python
+  socket.curve_secretkey = secret
+  socket.curve_publickey = public
+  socket.curve_serverkey = public
+  ```
+
+- The `heartbeat` client instead generates a fresh ephemeral keypair per socket and only needs the server's public key (`curve_serverkey`); it never receives the secret key. The heartbeat runs in its own ZMQ context and thread (so the GIL cannot stall it), and the curve options are therefore applied when its socket is created inside that thread.
+
+This "simplest version" (a single keypair used by both ends) choice from the variants considered in the pre-proposal is intentional. The connection file already carries the keypair, so reusing it avoids introducing a second key-distribution mechanism. `curve_serverkey` is what actually authenticates the connection; the rest configures encryption.
```
minrk:
```suggestion
This "simplest version" (a single keypair used by both ends) choice from the variants considered in the pre-proposal is intentional. The connection file already carries the keypair, so reusing it avoids introducing a second key-distribution mechanism. `curve_serverkey` authenticates the connection; the rest configures encryption.
```

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,226 @@
...
+
+### The `transport_encryption` policy
+
+The `transport_encryption` setting is configurable on the kernel manager in both `jupyter_client` and `jupyter_server`, with the values `disabled` (the default), `auto`, and `required`. `jupyter_server` is a thin layer: when the policy is not `disabled`, it forwards the policy into the per-kernel launch; the actual key generation, kernelspec check, and connection-file writing happen in `jupyter_client`. Setting the policy to `auto` or `required` while the local `libzmq` lacks Curve support (`zmq.has("curve")` is false) is rejected at configuration time, so the misconfiguration surfaces immediately rather than at first kernel start.
+
+The kernel itself has no encryption toggle of its own; it is purely _data-driven_. It applies Curve options if and only if the connection file it loads contains _both_ `curve_publickey` and `curve_secretkey`. A file with only one of the two is treated as unencrypted.
```
minrk: > A file with only one of the two is treated as unencrypted.

Given that this is security-related, we should probably treat this invalid partial config situation as a fatal error, rather than running insecurely.

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Kernelspec capability advertisement
+
+A kernel declares support with `metadata.supported_encryption`. The value is matched case-insensitively and may be either the string `"curve"` or a list containing `"curve"` (leaving room to advertise multiple schemes in future). The check happens _before_ the connection is established: capability cannot be discovered at runtime via `kernel_info_reply`, because by the time a reply could arrive the (unencrypted) connection already exists. This mirrors how JEP 66 gates the handshake pattern on the static `kernel_protocol_version` field.
```
minrk: Does it make sense to start with a list, rather than singleton string? Then we don't need to handle a change in type when a second item is added?

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,226 @@
...
+}
+```
+
+Both keys are optional and MUST appear together; their absence means no transport encryption, which is the legacy behavior. This would bump the schema to a new versioned `$id` (for example `connectionfile-v1.1`).
+
+### Backward compatibility and corner cases
+
+- _Default off._ With `transport_encryption` set to `disabled`, no keys are generated, the kernel launch is identical to today, and the Curve fields are absent from the connection file. A kernel that loads such a file behaves exactly as before.
+- `auto` with a non-advertising kernelspec silently starts the kernel unencrypted; `required` with such a kernelspec fails at startup.
+- _Transport._ Encryption applies to `tcp` only; `ipc://` connections rely on filesystem permissions as today, and `required` with a non-`tcp` transport is an error.
```
minrk: I'm not sure required + non-tcp should lead to an error.

1. I think IPC might be _more_ secure than curve in a number of situations, so we don't want secure config to be mutually exclusive, and
2. you can actually still use curve on other transports. It might be overkill, but it's fine

So I think it makes the most sense to either:

1. _ignore_ transport encryption on ipc, or
2. apply it the same as tcp

both of which make good sense and don't lead to errors.

minrk (commented):
Nice writeup and thanks for the first implementation!

145-transport-encryption/transport-encryption.md
krassowski:
```suggestion
The kernel itself has no encryption toggle of its own; it is purely _data-driven_.
It should apply Curve options if the connection file it loads contains _both_ `curve_publickey` and `curve_secretkey`.
A file with only one of the two should lead to an error and the kernel refusing to connect.
```

Carreau: Not a request for a change, but should we mention somewhere that kernels might at some point in the future refuse to start if there is not keys present in the file ?
I'm assuming we do want to have tcp+encytption be the default, and have no-encryption forbiden (unless you are debugging and actually go out of your ways to disable it)

krassowski: > Not a request for a change, but should we mention somewhere that kernels might at some point in the future refuse to start if there is not keys present in the file ?

I think this would be very sensible for some kernel wrappers designed to run in environments with sensitive data. The JEP already calls mandatory encryption out in "future possibilities", broadly saying we can consider granular enforcement:

> - _Mandatory encryption_ for remote or multi-tenant deployments, with per-channel or per-deployment policies once the remote story above is settled.

I will add another bullet point saying that kernels may in future refuse connection if these keys are not present.

145-transport-encryption/transport-encryption.md
Carreau: I think I would encrypt on IPC just to avoid extra branches in the code and treat the transport as opaque. I'm also thinking that there might be permission issues or misconfiguration that make an IPC a potential way to peek at data.

145-transport-encryption/transport-encryption.md
krassowski: Done in e1d495f

145-transport-encryption/transport-encryption.md
krassowski: Done in 0451cdd abd dc4e8e7

145-transport-encryption/transport-encryption.md
krassowski: Done in 528c9e1. I will follow with PRs to update PoC implementations. Since this is additive, we do not need to wait with releases.

krassowski:
@fleming79 would you like to be listed as co-author of the JEP? I would like to recognise your contribution with the restart fix and help with early testing.

fleming79:
Thanks for offering, but that's not necessary.

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,229 @@
...
+### Out of scope
+
+Considered out of scope for this JEP (addressable independently later):
+
+- _Remote and gateway kernels._ The server toggle is not propagated to gateway-managed kernels, and remote provisioners that ship the connection file over the network introduce a key-bootstrapping problem of their own. How encryption composes with remote kernels and with the JEP 66 registration-socket handshake is left to follow-up work.
+- _Key rotation._ Keys are fixed for the lifetime of a connection file (including across restarts); a rotation mechanism is out of scope here.
+
+## Future possibilities
+
+- _Handshake key exchange._ Rather than writing the secret into a shared connection file, encryption keys could be exchanged over the registration socket during the JEP 66 handshake, which would also make remote key exchange tractable without persisting secrets to disk.
```
JohanMabille: Should we handle the handshake pattern in this JEP? The keys should be communicated via the initial connection file (the one passing the registration socket info), since the kernel would bind its socket before communicating them back to the servef / client.

krassowski: _How_ would like to handle it in the JEP? Add another `curve-handshake` option and let kernels decide what they support? Change the default to use handshake pattern?

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,229 @@
...
+
+A kernel advertises that it understands encryption through a new key in its `kernel.json` metadata:
+
+```json
+{
+  "argv": ["...", "{connection_file}"],
+  "display_name": "Python 3 (ipykernel)",
+  "language": "python",
+  "metadata": {
+    "supported_encryption": ["curve"]
```
Carreau, line 37: thinking at loud; not requests to change.

Should this be plural ?

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Motivation
+
+With the default TCP transport non-authenticated clients with access to ports on the same machine are currently able to monitor IOPub, accessing any secrets exchanged during kernel runtime (code, outputs, etc). It can be only prevented by using IPC transport and file permissions.
+
+## Guide-level explanation
+
+The five ZMQ channels a kernel exposes (`shell`, `iopub`, `stdin`, `control`, and `heartbeat`) can now be wrapped in _transport encryption_: the connection between a client and a kernel is encrypted and mutually authenticated at the socket level, using ZeroMQ's [CurveZMQ](https://rfc.zeromq.org/spec/26/) protocol. CurveZMQ provides forward secrecy (it negotiates ephemeral per-connection session keys, so traffic captured today cannot be decrypted later even if the long-term keys leak) and connection-level authentication (a peer that does not hold the right key never gets its messages delivered, instead of being filtered out after the fact by an HMAC check).
```
Carreau: what do you mean by mutual ? I'm not sure I understand if the client distribute the keys, it can't distinguish between many servers (kernels) using the same keys , woudl that be with ZAP.

(also personal preference I prefer hardwrap to not point to full paragraphs)

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,229 @@
...
+- Client, front-end, and server authors generally get this for free through `jupyter_client` ≥ 8.9 and `jupyter_server` ≥ 2.20; the manager generates the keys and writes them into the connection file.
+- Operators flip one setting; the `auto` versus `required` distinction lets them roll encryption out gradually and then enforce it.
+
+There are two new diagnostics a user may notice. When a kernel runs over TCP _without_ encryption, `ipykernel` now logs a warning:
+
+> Kernel is running over TCP without encryption. All traffic, including any secrets in cell outputs, is readable by other processes on this machine. Use the IPC transport or start the kernel through a manager that provisions CurveZMQ keys.
+
+And requesting `transport_encryption` of `required` for a kernel that does not advertise support fails at startup:
+
+> RuntimeError: transport_encryption='required' but kernelspec does not declare `metadata.supported_encryption='curve'`.
```
Carreau: Accurate, but I think the message should be updated to say 'curve' not in ..., 

also supported_encryption maybe should have been plural, but that ship might have sailed now. See above comment.

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,229 @@
...
+issue-number: 75
+pr-number: 145
+date-started: 2026-06-10
+---
+
+# Transport encryption for ZMQ communication
+
+## Summary
+
+We propose to support encrypting the traffic exchanged between the kernels and client, adding a dedicated stanza in `kernelspec` to enable kernels to communicate support for encryption, adding a new set of fields for key exhange in the connection file, and using CurveZMQ protocol as a default implementation of such an encryption.
```
Carreau:
```suggestion
We propose to support encrypting the traffic exchanged between the kernels and client, adding a dedicated stanza in `kernelspec` to enable kernels to communicate support for encryption, adding a new set of fields for key exchange in the connection file, and using CurveZMQ protocol as a default implementation of such an encryption.
```

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Unresolved questions
+
+To be discussed during the JEP process:
+
+### Key isolation
+
+The reference implementation reuses one keypair for both the kernel's server identity and the local clients' identity, so the _secret_ key sits in the connection file and is loaded by every local client. This matches the `session.key` trust model and gives forward secrecy against passive capture, but the long-term secret in the file is sufficient for a live man-in-the-middle. Should we instead distribute only the server _public_ key and have clients (and possibly the kernel) use distinct keypairs? The pre-proposal raised holding the secret in an environment variable (for example `$JUPYTER_CURVE_SECRETKEY`) so the connection file alone is no longer enough to impersonate either end. As of now this JEP leaves standardizing it to a follow-up work, but should it be deemded universally desirable, we could support alternative key distribution mechanisms.
```
Carreau:
```suggestion
The reference implementation reuses one keypair for both the kernel's server identity and the local clients' identity, so the _secret_ key sits in the connection file and is loaded by every local client. This matches the `session.key` trust model and gives forward secrecy against passive capture, but the long-term secret in the file is sufficient for a live man-in-the-middle. Should we instead distribute only the server _public_ key and have clients (and possibly the kernel) use distinct keypairs? The pre-proposal raised holding the secret in an environment variable (for example `$JUPYTER_CURVE_SECRETKEY`) so the connection file alone is no longer enough to impersonate either end. As of now this JEP leaves standardizing it to a follow-up work, but should it be deemed universally desirable, we could support alternative key distribution mechanisms.
```

145-transport-encryption/transport-encryption.md
```diff
@@ -0,0 +1,229 @@
...
+
+### HMAC message signing
+
+Connection-level authentication can in principle remove the need for per-message HMAC verification, but the current implementation keeps signing as-is. For now we propose to keep it as defense-in-depth, but one could argue that opt-out of HMAC should be a part of this proposal to fully realise the promised benefits. While this is not currently part of the submitted PoCs, this JEP should encourage power users to experiment with opt-out, which may reduce the overhead of execution.
+
+### Connection-file schema
+
+As of now the added fileds are `curve_publickey` and `curve_secretkey` but we could consider a dedicated `encryption` group before we codify that in the schema.
```
Carreau:
```suggestion
As of now the added fields are `curve_publickey` and `curve_secretkey` but we could consider a dedicated `encryption` group before we codify that in the schema.
```

145-transport-encryption/transport-encryption.md
krassowski:
```suggestion
> RuntimeError: transport_encryption='required' but kernelspec does not declare 'curve' in `metadata.supported_encryption`.
```

145-transport-encryption/transport-encryption.md
krassowski: Addressed in 6a1ecce.

145-transport-encryption/transport-encryption.md
krassowski: IMO `encryptions` sounds funny in English. See https://english.stackexchange.com/questions/260385/is-it-ever-acceptable-to-pluralize-the-word-encryption-to-encryptions. If we were pedantic, we would need to change it to `supported_encryption_schemas` or so which is a bit long. I think it is ok to keep as-is.

145-transport-encryption/transport-encryption.md
JohanMabille: I think it should be very similar to the usual connection pattern; the kernel advertise what it supports in its kernelspec, then if the client decides to initiate a handshake, and if the kernel supports Curve encryption, the client passes the keys in the registration file.

krassowski: I think there are three approaches:
- a) the "initial connection info" that is used to communicate where the registration socket is could contain the curve keys (i.e. for local provisioner they get persisted on disk) - no change needed in this JEP maybe beyond a single sentence clarification - it is trivial and gives us no advantage
- b) the curve keys are sent over the registration socket - this way we do not need to persist the curve on disk; the downside is we would then also have a separate set of short-lived keys for registration socket so that one cannot eavesdrop on that connection either
- c) we move the key generation to kernels; the kernel generates the curve key pair and sends _only_ the public key thorough the registration socket; this is possibly the best future state (and my initial PoC included kernels generating the keys) but has a steeper cost for adoption: kernels need to generate their own keys (so be built with libsodium support for zmq), we need to have JEP 66 implemented etc.

The line 244 is saying "in future we may want to consider (b) or (c)". Are you suggesting that for now we just document (a)?

145-transport-encryption/transport-encryption.md
Carreau: Fair. With a pure linguistic view it make sens.

145-transport-encryption/transport-encryption.md
minrk: I think it's fine to leave handshake/registration as future work. If folks think it needs to be addressed here, we should use a curve key _on the registration socket itself_, since registration in the current handshake suffers from the same issues motivating curve in the first place.

JohanMabille: > The line 244 is saying "in future we may want to consider (b) or (c)". Are you suggesting that for now we just document (a)?

Yes, I think (a) can be part of this JEP,  and (b) and (c) left as possible future work.

krassowski: I pushed 4ad2d85883d4d9cc09ea20d02294f6b666d90275 let me know if I missed something in there. Otherwise, I think this is ready for a vote.

JohanMabille: I think this is ready for a vote.

krassowski:
Can the vote here proceed or is it blocked?

JohanMabille:
I was waiting for the Jupyter Accessibility subproject to elect a new SSC representative before triggering the vote, but we can proceed before if that's something urgently requested.

wshanks:
Would jupyter_client ever change its default for `transport_encryption` from `disabled` to `auto`? With https://github.com/ipython/ipykernel/pull/1515, I started seeing a lot of the "Kernel is running over TCP without encryption" warnings and see that that even came up in the PR (as a problem for nbclient). In my case, the warnings come from kernels run by jupyter-sphinx. I wasn't sure if this were something that a project like jupyter-sphinx should deal with itself or if the lower level packages would work it out.

(Sorry if there is a better place to ask).

krassowski:
I think we should switch to `auto` in the future. I am not sure when but since this adds additional keys to the connection file it is not really side-effect free, so it makes sense to wait until this feature is well established.

> I started seeing a lot of the "Kernel is running over TCP without encryption" warnings

Well, this is just a warning to protect users. Maybe we should tighten when that warning is shown - for example it adds little value when running on CI. Would that help?

wshanks:
Maybe tightening up when the warning is shown would help but I care about running locally as well as CI. In this case, the context is just when building docs. I worked out how to turn on encryption to avoid the warning in my case in https://github.com/qiskit-community/qiskit-experiments/pull/1654, but I was wondering if jupyter-sphinx should do something like that itself.

PR #147, opened 2026-07-17
Convert our docs site to MyST

choldgraf:
This converts our website infrastructure to use MyST, with the following goals:

- Use the more up-to-date docs infrastructure we're rolling out across Jupyter
- Improve our table of JEPs to be searchable and easier to maintain
- Put JEP information in the metadata of each JEP and re-use that, rather than hand-maintaining a second table in the readme

[😎 Build preview here](https://jupyter-enhancement-proposals--147.org.readthedocs.build/en/147/)

CleanShot 2026-07-17 at 13 57 05@2x


westurner:
YAML frontmatter schema;

- `name` instead of `title`
- `authors` could be a list
- `date` -> `dateModified`?
  - is `date` `dateModified`?
- .
- Thing > schema.org/CreativeWork > . > Jupyter:JEP
  - Thing has properties like: 
    - `name` instead of `title` (Thing.name)
    - `description` (Thing.description)
  - CreativeWork has properties like:
    - `author`,
    - `funder`, `funding`, `sponsor`
    - .
    - `dateCreated`
    - `dateModified`
    - `datePublished`?
    - .
    - `publication` <URI/URL for all JEPs>
    - `partOf` / `hasPart`
    - .
    - `workTranslation` / `translationOfWork`
    - .
    - `version`
- https://schema.org/CreativeWork
- W3C YAML-LD
  - later transform frontmatter metadata into JSON-LD,
    and then apply an @context
  - or the schema is already YAML-LD with a @context
- add jsonschema (and/or W3C SHACL) validation in the actions build
  - frontmatter data validation should cause build failure

westurner:
The OKF LLM wiki spec also specifies a markdown YAML frontmatter schema;

From HN the other day;

> It looks like a revision to OKF format could include: support for RDF URIs as labels, schema.org RDFS alignment (:name @en, :about, :description,)

>> You can think of our app as a general purpose IDE for any knowledge base/wiki/markdown.
>>  
>> Docs:  https://openknowledge.ai/docs/workflows/supporting-open-knowledge-format

An LLM wiki spec for frontmatter in JEPS would enable DRY indexing and GraphRAG search,

Linked Data is still lower-cost than LLMs,

So justified then, YAML-LD frontmatter (and with schema.org classes and properties)

- Open Knowledge Format (OKF)
  https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md

westurner:
FWIW here's a v0.0.2 myst-md-rs and sphinxdocrs (and jinja2rs with minijinja,  optionally with the Python compat layer) 
https://github.com/westurner/dsport/tree/main/src/myst-md-rs ;

> Rust port of MyST-Parser, built on top of pulldown-cmark.


- this already needs work, and YAML-LD OKF support:
  https://github.com/westurner/dsport/blob/main/src/myst-md-rs/src/frontmatter.rs

Jupyter-book v1 and v2 (Rust) and quarto (pandoc) could add support for (YAML-LD) OKF frontmatter too

myst.yml
```diff
@@ -0,0 +1,31 @@
...
+    - title: JEPs
+      children:
+        - pattern: "**/*.md"
+site:
+  template: book-theme
+  options:
+    logo_text: Jupyter Enhancement Proposals
+    logo: https://github.com/jupyter/governance/raw/main/docs/images/jupyter-logo-light.png
+    logo_dark: https://github.com/jupyter/governance/raw/main/docs/images/jupyter-logo-dark.png
+    analytics_plausible: jupyter.org
```
jasongrout, line 31: Plausible has deprecated the data-domain method, and instead these days recommends using their per-domain url. I did something like this in https://github.com/jupyter/governance/pull/296, for example.

jasongrout (commented):
I think the table rendering really looks nice!

myst.yml
choldgraf: Ahh - that might need an upstream fix...let me look into that

choldgraf: I've got an issue about that here:

- https://github.com/jupyter-book/myst-theme/issues/932

but it looks non-trivial to fix, and it also seems like the legacy plausible code will work for a bit. What do you think about using the legacy script here and then upgrading once it's supported upstream? Or would we lose something important @jasongrout ?

ps: for governance we have the same challenge because I'm pretty sure we upgraded that to myst, and might have messed up the config when we did it

jasongrout (approved):
LGTM. We can fix the plausible update at the myst level, then port back to here. Since this is still draft, I won't merge it, but +1 from me. Thanks for making this better!

122-jupyter-book-incorporation/jupyter-book-incorporation.md
```diff
@@ -144,6 +148,95 @@ From a technology standpoint, we are adopting tools that have a good accessibili
 
 (note the "last updated" field is out of date, and reflects when this draft first started in early March 2024)
 
-```{csv-table} 
-:file: repositoriestoinclude.csv
+```{csv-table}
+:header-rows: 1
```
choldgraf: I think this could just use an `{include}` directive per MyST's instructions?

12-jupyter-kernel-gateway-incorporation/jupyter-kernel-gateway-incorporation.md
```diff
@@ -1,3 +1,10 @@
+---
+JEP: 12
```
choldgraf, line 2: Is there a more reliable way we could get the JEP number without requiring it to be hard-coded? If not then this is probably OK, but for example if JEP number is always the same as issue number maybe we could re-use that?

18-jupyter-declarativewidgets-incorporation/jupyter-declarativewidgets-extension-incorporation.md
```diff
@@ -174,8 +181,11 @@ The declarativewidgets extension is packaged using setuptools, released on PyPI,
 @lbustelo
 
 ---
-<a name="1"></a> [1] R support though [IRkernel](https://github.com/IRkernel/IRkernel)
+(1)=
```
choldgraf: We should try and turn these into normal footnotes in the markdown doc since myst supports footnote syntax like `[^1]` and `[^1]:`

conf.py
```diff
@@ -1,32 +0,0 @@
-author = "Project Jupyter"
-copyright = "2021"
-exclude_patterns = ["_build", ".nox", ".github"]
-extensions = [
-    "myst_parser",
-    "sphinx_copybutton",
-    "sphinx_external_toc",
-    "sphinx.ext.intersphinx",
-    "sphinx_book_theme",
-    "sphinxcontrib.mermaid",
```
choldgraf, line 10: double check that our mermaid diagrams still work (if used)

withdrawn-jeps.yml
```diff
@@ -0,0 +1,10 @@
+# JEPs that were never merged; each url points at its pull request.
+# Rendered by the {listing} directive in README.md.
+- {JEP: 4, status: Withdrawn, title: New Notebook Format for improved workflow integration, url: "https://github.com/jupyter/enhancement-proposals/pull/4"}
```
choldgraf: maybe this should be expanded YML instead of this short-hand for easier human readability

choldgraf (commented):
gave it a little review and here are a few things I noticed

12-jupyter-kernel-gateway-incorporation/jupyter-kernel-gateway-incorporation.md
jasongrout: From the [instructions](https://github.com/jupyter/enhancement-proposals/blob/master/jupyter-enhancement-proposal-guidelines/jupyter-enhancement-proposal-guidelines.md#phase-2-request-for-comments-for-the-jep):

> Phase two begins when the JEP author submits a draft of the JEP as a PR to the JEP repository. The Shepherd assigns a number to the JEP according to the pull-request number, and updates the README of the JEP repository with information about the now in-progress JEP. The Shepherd then works with the JEP author to identify and notify relevant individuals in the Jupyter community to review and comment.

So it sounds like we can make the assumption that the JEP number is tied to the PR number right now. However, I think it would be better if we instead chose the JEP number sequentially, like in Python PEPs - the JEP is a permanent artifact, whereas the PR number is an implementation detail of how we created that artifact on GitHub. I don't know if the number system is likely to change in the future, though, and we can always adapt the metadata if we did change.

18-jupyter-declarativewidgets-incorporation/jupyter-declarativewidgets-extension-incorporation.md
jasongrout: I went ahead and pushed a change to do this.

choldgraf:
OK I made another few edits to this and I think it's pretty safe to try now. @jasongrout wanna merge and we can spot-check more improvements as needed?

12-jupyter-kernel-gateway-incorporation/jupyter-kernel-gateway-incorporation.md
choldgraf: Yeah that makes sense - ok I'm happy to leave it for now, always better to keep things explicit anyway

jtpio:
FYI this may likely break existing links, as noticed in https://github.com/jupyterlab/jupyterlab/pull/19212 with a link to a JEP.

No big deal since it can easily be updated, just wanted to raise it here in case changing existing URLs was not intentional.

krassowski:
I just saw it when trying to open a JEP from Google it displayed a 404 page. I think we should fix the links, there are in too many places.

PR #151, opened 2026-07-29
Upload the built site as an artifact

jtpio:
This will help preview the site more easily without building it locally since it can be downloaded via the GitHub Actions page:

image

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,8 +31,17 @@ jobs:
         pip install nox
         nox -s docs
       env:
-        # The site is hosted at jupyter.org/enhancement-proposals (a subpath)
-        BASE_URL: /enhancement-proposals
+        # The deployed site is hosted at jupyter.org/enhancement-proposals (a
+        # subpath). Pull request builds are downloaded as an artifact and served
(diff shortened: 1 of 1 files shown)


choldgraf (approved):

choldgraf:
Lgtm - what do you think about trying out one of those PR preview link generator actions? That could get us easier PR preview links (I think they deploy to dedicated branches or something)

jtpio:
Maybe it could be useful, although iirc these setups feel a bit more like workarounds (since GitHub still does not support previews natively).

Another option could be to use ReadTheDocs for the previews, which has been working quite well for other projects. For example https://github.com/jupyter/try-jupyter uses GitHub Pages for the `main` deployment, and RTD for previews.

bollwyvl:
Yep, per-PR review sites with nice little badge comments are certainly
worth the one file and one time UI registration. Actually hosting something
there (e.g. `jep-history.rtfd.io`) with the potential for having
`{version}/{l10n}` urls is very useful, even if the canonical url does not
change.

choldgraf:
+1 on using read the docs or netlify too - I use both a lot and have good experiences with each

PR #154, opened 2026-09-15
Fix author lists

krassowski:
Another thing that I noted is how the author lists on most JEPs were not rendering correctly, this fixes it

### Before

image

### After

https://github.com/user-attachments/assets/8d28263b-aeb5-4fc9-845e-7f3caa660238


JohanMabille (approved):
Thanks!

